PR #20, opened 2020-05-07
docs: add description main module

redeboer:
Also removes unused graphvis package


expertsystem/__init__.py
```diff
@@ -1,13 +1,38 @@
-__all__ = ["amplitude", "state", "topology", "ui"]
+"""
+The expert system facilitates building am amplitude model.
```
spflueger, line 2: an

Issue #4, opened 2020-01-16
Isolate XML keys to a separate submodule


redeboer:
Currently, all modules of the expert system have some dependency on `xmltodict`, see for example [here](https://github.com/ComPWA/pycompwa/blob/39d8ecd8c1bd756184bd52495cd5b212e8bcf810/pycompwa/expertsystem/state/particle.py#L24). Would be better to rewrite this into general interfaces and have specifications for XML separated out.

spflueger:
As mentioned in #25, I'm missing the big picture. Separating the xml related things into a module is not bad, but I think we can even do one better.
I would like to have all xml (or other file format) related code to vanish completely.
Similar to the current code a single dict (with a defined structure) contains all of the information. This can be written or read from file in various file formats, as long as they are compatible with the dictionary structure.

redeboer:
Alright, let's move this discussion to a separate design document separate out more focused issues from that.

PR #21, opened 2020-05-07
test: clean up tests folder

redeboer:
There were still some references to `pycompwa`, they have been removed now.

* Add more pytest "slow" markers
* Remove "disabled" tests
* Move non-tests to examples folder
* Remove 'addopts' from `pytest.ini`, the `pytest` output was a bit too informative


redeboer:
> Looks good. We have to make sure to have some kind of tests like the angular dist in the higher level frameworks like pycompwa or tensorwaves though

Should be made into some issue in `tensorwaves`, I guess

redeboer:
Btw this PR is extracted from ongoing work on https://github.com/ComPWA/expertsystem/issues/4, to keep the eventual PR small

spflueger:
> > Looks good. We have to make sure to have some kind of tests like the angular dist in the higher level frameworks like pycompwa or tensorwaves though
> 
> Should be made into some issue in `tensorwaves`, I guess

yeah

PR #25, opened 2020-05-13
refactor: isolate xml tags to separate module

redeboer:
Closes https://github.com/ComPWA/expertsystem/issues/4

This is a simple first step towards isolating the XML tags to a different submodule. There are still classes like `HelicityAmplitudeGeneratorXML` that should become more general (and work with an XML parser argument, for example), but that's for a follow-up PR.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -78,6 +78,7 @@ primary_domain = "py"
 nitpicky = True  # warn if cross-references are missing
 nitpick_ignore = [
     ("py:class", "StateTransitionGraph"),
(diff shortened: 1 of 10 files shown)


spflueger:
Yeah, so I'm missing the general idea or concept. So I don't see how this and PR #31 will help to solve the actual problem. I will comment on issue #4 regarding this.
I would say we should come up with a concept document to get a clean picture of what is going wrong right now, how we want things to be like and what options we have. We can discuss the concept more in a related issue.

So I'm afraid until I see some concept to get the full picture of where we want to go with this, I cant make a meaningful review.

PR #36, opened 2020-05-26
feat: add yaml support for particle list

redeboer:
The `particle_list` now contains `dict`s instead of `OrderedDict`, so that it can be processed by `pyyaml`. The structure of `particle_list` is still the same though and does not yet reflect the [structure of `particle_list.yml`](https://github.com/ComPWA/expertsystem/blob/74e9bf34825155c7b044375f05e140e196b0e036/particle_list.yml).

### DIFF
--- a/expertsystem/state/particle.py
+++ b/expertsystem/state/particle.py
@@ -3,15 +3,18 @@ This module defines a particle as a collection of quantum numbers and
 things related to this
 """
 import logging
+from abc import ABC, abstractmethod
+from collections import OrderedDict
 from copy import deepcopy
 from enum import Enum
-from abc import ABC, abstractmethod
-from numpy import arange
 from itertools import permutations
 from json import loads, dumps
-from collections import OrderedDict
+
+from numpy import arange
 
 import xmltodict
+
+import yaml
 
 from ..topology.graph import (
     get_initial_state_edges,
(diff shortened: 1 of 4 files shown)


expertsystem/state/particle.py
```diff
@@ -40,6 +43,11 @@
 )
 
 
+def to_dict(input_ordered_dict: OrderedDict) -> dict:
+    """Convert nested `OrderedDict` to a nested `dict`."""
+    return loads(dumps(input_ordered_dict))
+
+
```
spflueger, line 50: Because this function is simply something needed for xml, I would plug this function into that scope.

redeboer: b89fbac754f5555de01412440c681dc84b721ea3

redeboer:
@spflueger the issue with this PR is probably not as much its content (that is quite trivial), but more the direction into which it is moving and whether the next step (i.e., replacing the source of truth for `particle_list` with the new `dict` structure as defined by [`particle_list.yml`](https://github.com/ComPWA/expertsystem/blob/74e9bf34825155c7b044375f05e140e196b0e036/particle_list.yml)) should be included in this PR.

expertsystem/state/particle.py
```diff
@@ -3,16 +3,19 @@
 things related to this
 """
 import logging
+from abc import ABC, abstractmethod
```
spflueger, line 6: why is this imported here?

spflueger: nvm

PR #41, opened 2020-05-29
ci: standardise VSCode settings

redeboer:
This may be a solution to some of the issues we had with mismatching configurations. Seems to be common practice among the VSCode team:
https://stackoverflow.com/a/32979933

@spflueger Feel free to add debug targets or so and additional settings.


.vscode/settings.json
```diff
@@ -0,0 +1,27 @@
...
+  ],
+  "python.linting.mypyEnabled": true,
+  "python.linting.pydocstyleArgs": [
+    "--match='(?!test_).*\\.py'"
+  ],
+  "python.linting.pydocstyleEnabled": true,
+  "python.linting.pylintCategorySeverity.refactor": "Information",
+  "python.linting.pylintEnabled": true,
+  "python.linting.pylintUseMinimalCheckers": false,
+  "python.pythonPath": "/home/remco/anaconda3/envs/expertsystem/bin/python",
```
spflueger, line 22: what to do about this?

redeboer: Oops... :man_facepalming: 81c996d1b8073bae25dc9c82d556909b3e0b9b65

spflueger: But how does this work if everyone defines this line himself (Because that's a workspace setting)? You have to ignore that line somehow. I don't quite understand how this will be handled nicely...

redeboer: Yes this is indeed a slightly problematic point... The way I solve it, is create a [(multi-root) workspace](https://code.visualstudio.com/docs/editor/multi-root-workspaces) for experstystem, so that you can specify the python path in the workspace settings and leave the folder settings unaffected.

spflueger:
> Yes this is indeed a slightly problematic point... The way I solve it, is create a [(multi-root) workspace](https://code.visualstudio.com/docs/editor/multi-root-workspaces) for experstystem, so that you can specify the python path in the workspace settings and leave the folder settings unaffected.

Note: if you open the repository in VSCode as a folder only, the settings will say 'workspace settings', but those are defined under `.vscode/settings.json` ('folder settings'). Once you save a workspace, a middle layer of settings becomes available, that stores its settings under a `*.code-workspace` file. 

All right sounds good

.vscode/settings.json
```diff
@@ -0,0 +1,34 @@
...
+    "--match='(?!test_).*\\.py'"
+  ],
+  "python.linting.pydocstyleEnabled": true,
+  "python.linting.pylintCategorySeverity.refactor": "Information",
+  "python.linting.pylintEnabled": true,
+  "python.linting.pylintUseMinimalCheckers": false,
+  "python.testing.pytestArgs": [ // https://github.com/microsoft/vscode-python/issues/693
+    "tests",
+    "--no-cov",
+    // "--cov-config=tests/.coveragerc",
```
spflueger, line 25: I would remove those comment lines, and add it once you want code coverage to be in there

.vscode/settings.json
spflueger: Ok, sounds good. Let's give this a try then, and in case we don't have a good experience with this we can change it. Maybe a default settings file would an option to keep in mind https://stackoverflow.com/a/48387809

Issue #22, opened 2020-05-08
Expand UI for AmplitudeModel building


redeboer:
Perhaps this issue is even more general: both `expertsystem` and `tensorwaves` (perhaps even `pycompwa`) need to agree on the syntax on the recipe, so it's natural to put some sort of recipe module into the expert system.

_Originally posted by @redeboer in https://github.com/ComPWA/tensorwaves/pull/60/review_comment/create_

redeboer:
@spflueger This is related to #40, as the recipe file contains a selection of the particle list that forms the starting point for the expertsystem.

Since the internals of the expertsystem only functions if the input particle info comes in the form of `dict`s that assume our old XML format (those `dict`s are used to populate the edges of the graph), I plan to 'enclose' the expertsystem input (namely a particle list) and output (a recipe file) in the form of data structures that can load and dump to those XML-formatted `dict`s.

A series of PRs will follow, starting with a `ParticleCollection` containing `Particle` instances that can be built from both YAML and XML and then dumps its info into a nested `dict`. Later on, this internal nested `dict` structure can then slowly be replaced `Particle` instances.

redeboer:
Intended structure/workflow:
![ExpertSystem-TensorWaves integration](https://user-images.githubusercontent.com/29308176/88162671-ecf1c780-cc11-11ea-8748-084de8f44918.png)
So the idea is that all recipe I/O *and* the recipe structure is pinned down in the form of classes. Then the expertsystem becomes an actual dependency of tensorwaves and there's no need to define `dict` to TF object conversions each time the recipe file structure changes.

redeboer:
Closed by #454

PR #42, opened 2020-06-02
feat: add data structures for particle info

### LINKED ISSUE #40: Encode particle info in the form of a class or data structure
Currently, particle info is assumed to be a dictionary, and `Enum`s are used to 'standardise' the `dict` keys across the frameworks.

As the new source of truth is to be the [`particle_list.yml`](https://github.com/ComPWA/expertsystem/blob/243c8c757cd8403d34ecbaafaf2126b269fb5824/particle_list.yml), which results in a different `dict` structure than that of the existing [`particle_list.xml`](https://github.com/ComPWA/expertsystem/blob/243c8c757cd8403d34ecbaafaf2126b269fb5824/particle_list.xml), we run into trouble. I therefore suggest it's better to encode the particle data structure in terms of a `class` rather than a `dict` and to remove the `Enum`s and `get_...` functions in `particle.py`. 

Related to #22, although I would treat the particle list separately, as the list is the starting point for the expert system. In that sense (the "source of truth"), this issue is related to #23 and #33.

redeboer:
Closes #40 

This PR is part of the ongoing effort to make the codebase of the expertsystem more readible.

- Cleans up and improves the existing `Spin` class
- Introduces a `ParticleDatabase` (dict of `Particle` object) that is to replace the `particle.DATABASE` global
- Since switching to `ParticleDatabase` requires refactoring almost the entire codebase, all old functionality has been moved to a namespace `particle.deprecated`, which will be gradually removed in follow-up PRs.
- Run mypy over tests as well (but do not enforce type hints in the definitions there)
- Compute finer test coverage with [branch coverage](https://coverage.readthedocs.io/en/latest/branch.html).

*(This PR has been rebased on #41)*


particle_list.yml
```diff
@@ -1,821 +1,780 @@
 ParticleList:
   gamma:
     PID: 22
-    Mass: { Value: 0.0 }
+    Mass: 0
```
spflueger, line 4: I think we have to think about this to be sure we do the right thing here. Having this dict structure with `Value`, `Error` etc could help make it easier to create Parameter objects from this. On the other hand, Error does not make much sense here in terms of setting up a fit (minuit uses this for initial step sizes but its not necessary).

particle_list.yml
```diff
@@ -1,821 +1,780 @@
...
       Parity: -1
       Cparity: 1
       Gparity: -1
-      IsoSpin: { Value: 1, Projection: 0 }
-      BaryonNumber: 0
-      Charm: 0
-      Strangeness: 0
-
+      IsoSpin:
+        Value: 1
```
spflueger, line 135: Magnitude

expertsystem/state/particle/__init__.py
```diff
@@ -0,0 +1,437 @@
...
+    ln_electron: int = 0
+    ln_muon: int = 0
+    ln_tau: int = 0
+
+    @property
+    def has_width(self) -> bool:
+        return self.width is not None
+
+
+class ParticleDatabase:
```
spflueger, line 151: This looks quite big for what it can do. I have to take a closer look at this

redeboer:
The DeepCode error is incorrect, because `pyyaml` and `xmltodict` don't work with binary mode. It seems the only way to ignore this issue is to put deepcode comments into the code. I don't like that, so I suggests giving up the deepcode checks -- it's too inflexible and and we already have sufficient checks through Codacy.

particle_list.yml
redeboer: In the current implementation, you can do both so that the YAML file becomes readable. That's also one of the reasons for https://github.com/ComPWA/expertsystem/pull/42#discussion_r433749950

expertsystem/state/particle/__init__.py
redeboer: Yes, the 'problem' is the functionality to load from both XML and YAML (and, worse, the existing XML structure). Technically, this class contains several factories to construct `Particle` objects and the objects it contains from `dict` structures, such as `load_xml`. The interface doesn't reveal that because these factories are wrapped as inline functions or private methods in case the dict structure is the comparable in the case of both XML and YAML.

particle_list.yml
redeboer: I agree that Magnitude would be a better name, but I've tried not to change the expected structure of the YAML file (i.e. names of the keys and such). So, strictly speaking, I would separate this suggestion into a new issue, so that this PR only pertains to the introduction of a `ParticleDatabase` that can read from the existing `particle_list.yml` and `particle_list.xml` files.

expertsystem/state/particle/__init__.py
redeboer: I start to think this problem is strongly related to #22. In the end, this module should just serve as a collection of data containers for particle info -- the `load` and `write` methods are related to serialisation to a custom XML or YAML format that not only contains particle info, but also the intensity recipe etc.

So what I'm currently giving a try is to design some sort of file adapter or recipe module that can be easily expanded, but that for now would only serve as a factory for the `ParticleDatabase` class. This module should also easily be accesible usable by an external package (i.e. Tensorwaves).

redeboer:
https://github.com/ComPWA/expertsystem/pull/42#discussion_r433749950
All serialisation functionality has been removed so that the classes in this are purely data objects. The construction of these data objects will be further addressed in #44

expertsystem/state/particle/__init__.py
redeboer: Okay, let's move this discussion to #44, it's a bigger design decision. The current PR should only fix #42, so it was a mistake to start working on `load`/`write` functions in this PR already.

redeboer:
Note that we may also want to add some tools that allow you to create a new particle instance based on another particle instance as template in Python code (so that you can for instance create duplicates of fake particles such as an e+i- collision at different energies or hypothetical particles). But as that functionality is not yet needed, I did not spend time on creating an interface for this yet.

redeboer:
This PR will be split up into small PRs that address specific issues

PR #48, opened 2020-06-05
adr: initialize architectual design records

### LINKED ISSUE #47: Create template for Architecture Decision Records
Some nice resources on ADRs:
- [Homepage of the ADR GitHub organization](https://adr.github.io/)
- [SATURN 2017 Talk: Architecture Decision Records in Action](https://www.youtube.com/watch?v=41NVge3_cYo)
- [Markdown Architectural Decision](https://adr.github.io/madr/)

Probably best to follow the structure of the [adr/madr](https://github.com/adr/madr) repo.

spflueger:
Closes #47


doc/adr/0000-use-markdown-architectural-decision-records.md
```diff
@@ -0,0 +1,26 @@
+# Use Markdown Architectural Decision Records
```
redeboer, line 1: Do we need to have an ADR for this PR?

spflueger: I think it makes sense to start off this way. One thing we should think about still is wether to add [ADR-xxxx] into the title or let sphinx do the numbering based on the filenames. I almost then to sphinx numbering because it avoids typos.

doc/adr/template.md
```diff
@@ -0,0 +1,72 @@
...
+
+* Status: [proposed | rejected | accepted | deprecated | … | superseded by [ADR-0005](0005-example.md)] <!-- optional -->
+* Deciders: [list everyone involved in the decision] <!-- optional -->
+* Date: [YYYY-MM-DD when the decision was last updated] <!-- optional -->
+
+Technical Story: [description | ticket/issue URL] <!-- optional -->
+
+## Context and Problem Statement
+
+[Describe the context and problem statement, e.g., in free form using two to three sentences. You may want to articulate the problem in form of a question.]
```
redeboer, line 11: I wonder if we should somehow enforce hard-wrapping lines. Thing is, new ADRs are presented in the form of a PR and there you mostly review the files through a diff view.

doc/adr/0000-use-markdown-architectural-decision-records.md
redeboer: Yes I agree with the ADR-xxxx. Downside: it has to be done manually each time. Upside: the number is hardcoded in the document, so hard to temper with.

redeboer: In that case we should indeed remove the sphinx numbering in the toc table

doc/adr/template.md
spflueger: Hmm, but the files wont be touched after the PR is merged. So it would only help during the PR reviewing phase, and there the whole document is completely new anyways...

redeboer: Well, I already have the problem atm reading this line, imagine an entire paragraph. Diff viewers don't soft wrap lines.

redeboer: Maybe prettier is a solution? Seems they are thinking about this problem specifically in the case of MarkDown
https://prettier.io/docs/en/options.html#prose-wrap

doc/adr/0000-use-markdown-architectural-decision-records.md
redeboer: Personally, I use the [Rewrap extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) in VScode. Note that we already have a setting specified for that extension here:
https://github.com/ComPWA/expertsystem/blob/cf1728eedf466f2d590a6eee77ddfb2b94359b76/.vscode/settings.json#L29
The question is how to enforce this in a more general way, through a pre-commit hook for instance.

doc/adr/template.md
spflueger: I guess it does not really hurt, but how to enforce it? We can just promote it and recommend tools. But that would be ok for me

doc/adr/template.md
```diff
@@ -0,0 +1,72 @@
+# [short title of solved problem and solution]
+
+* Status: [proposed | rejected | accepted | deprecated | … | superseded by [ADR-0005](0005-example.md)] <!-- optional -->
+* Deciders: [list everyone involved in the decision] <!-- optional -->
+* Date: [YYYY-MM-DD when the decision was last updated] <!-- optional -->
```
spflueger, line 5: Should be make those things required?

redeboer: I think the date field is redundant for a file under source control

doc/adr/template.md
```diff
@@ -0,0 +1,72 @@
+# [short title of solved problem and solution]
+
+* Status: [proposed | rejected | accepted | deprecated | … | superseded by [ADR-0005](0005-example.md)] <!-- optional -->
+* Deciders: [list everyone involved in the decision] <!-- optional -->
```
redeboer, line 4: Even this: why spend time on documenting it if it's already embeded in the commits (and in that extend the PR number through which the commit was merged into the master)

doc/adr/0000-use-markdown-architectural-decision-records.md
redeboer: Okay I think we should forget about line wrapping.
1. Too much effort to enforce this, even if there are tools available
2. GitHub offers several [diff view options](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-comparing-branches-in-pull-requests#diff-view-options), so there are plenty of ways to conveniently review an ADR that does not have wrapped lines.
3. PR review in VScode does not have to be done through a diff editor, so you have soft wrapping at your disposal there.

doc/adr/template.md
spflueger: Hmm, but not everyone will commit. Ok the PR will show more info. I guess optional would be fine there

doc/adr/template.md
spflueger: True

spflueger: I think it would be good if there would be a quick link to the PR though

redeboer: But the link to the PR will be available through the eventual squashed commit and then in the PR you can see who drove the decision.

doc/adr/template.md
spflueger: I added deciders as optional again

doc/adr/template.md
```diff
@@ -0,0 +1,62 @@
+# [ADR-xxxx] <!-- short title of solved problem and solution -->
+
+Status: **[proposed | rejected | accepted | deprecated | … | superseded by [ADR-xxxx](xxxx-example.md)]**
+Deciders: [list everyone involved in the decision] <!-- optional -->
```
redeboer, line 4: There's a slight problem here with regard to Sphinx, I'm afraid: Sphinx does not render single linebreaks as a linebreak. So we should either have two linebreaks here or use bullet points.

PR #61, opened 2020-06-16
ci: run pytest from root + add py38 job

redeboer:
Also run `pytest` from the main directory, so better to treat as a separate PR.


redeboer:
The important part of this PR is the coverage: it didn't change
https://github.com/ComPWA/expertsystem/pull/61#issuecomment-644759774
I point it out because I remember we had problems with this when we ran `pytest` from the main directory in pycompwa.

PR #69, opened 2020-06-18
build: bump version to 0.1.1-alpha

redeboer:
Made a mistake with the version in #59... I thought to have [0.1.1](https://github.com/ComPWA/expertsystem/milestone/7) implement YAML for the helicity formalism (so it can be used in tensorwaves), while [0.1.2](https://github.com/ComPWA/expertsystem/milestone/8) implements the canonical formalism.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ def long_description():
 
 setuptools.setup(
     name="expertsystem",
-    version="0.1-alpha",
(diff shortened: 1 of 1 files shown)


spflueger:
I'm not even sure this is a mistake. Seems like good practice to me if the version bumping is done in a separate PR, or?

Issue #68, opened 2020-06-17
Standardize and automate release flow


redeboer:
## Problem description

Have had a few times that pushing a tag results in a failed deploy job, e.g. https://travis-ci.com/github/ComPWA/expertsystem/jobs/350476070, because I forgot to bump the version in the `setup.py` file. There are better tools for that, like [`bumpversion`](https://pypi.org/project/bumpversion/).

Also, we haven't really thought about how to deal with releases. Linear commit history already provides some sort of clean versioning, now it's just a matter of summarising the most important changes for each release. For now, I follow the [release style that tensorflow uses](https://github.com/tensorflow/tensorflow/releases). Although we're unclear about [Semantic Versioning](https://semver.org/).

## Summary
* Should we use [Semantic Versioning](https://semver.org/)?
* How to decide about whether a release is a patch or a minor version update?
* How to make that as effortless (automatic) as possible?
* Which (https://github.com/ComPWA/expertsystem/pull/70) conventions to use for the release draft?
* How to keep Git tags in sync with the (single source of truth)
* Should we reconsider our packaging tool/setup (i.e.`setup.cfg`, `pyproject.toml`, ...)

This differs from https://github.com/ComPWA/expertsystem/pull/70 in the sense that in this repository we want to try out and illustrate some tools for a nice release flow. In the PWA Pages we describe that release flow after it turns out to work well. The PWA Pages then serve more as a guideline.

## Some resources
- https://packaging.python.org/guides/single-sourcing-package-version/
- https://semver.org/
- https://stackoverflow.com/a/5582128 (on Git and single source of truth for package version)

spflueger:
Regarding the version of setup.py: The current situation is not ideal, but its fairly ok. Not sure if bumpversion is foolproof. Some kind of hook in github would almost be best, that has access to the tag stage and will deny a tag if the version does not match. Don't know what state of the art is here.

Regarding the versioning: Semantic versioning is nice and I would love to do that (to get some hands on experience with this). But for me that would imply that you keep several major version branches that are maintained simultaneously. So I think that will add to much workload.

spflueger:
Lets turn this into an ADR, to keep our decision fixed. I think this should almost be ComPWA global, decision...

redeboer:
If it's global, imho the best place would be
https://github.com/ComPWA/PWA-pages
[...]

PR #70, opened 2020-06-18
adr: choose packaging system and versioning scheme

redeboer:
Initiated from issue #68 

Some things we should address:
- [x] How to keep in sync with tags/releases on GitHub?
- [x] Decide about packaging system (`pyproject.toml`, `setup.cfg`, and what have you...)
  --> *Have a look at [PEP 518](https://www.python.org/dev/peps/pep-0518/)*

Note that this ADR differs from https://github.com/ComPWA/PWA-pages/issues/7 in that this ADR sets out tools (i.e. 'locally' for the expertsystem repo), while the one on the PWA Pages will (as a follow-up) formulate what that release flow looks like (i.e. 'globally' for all sub-projects).


redeboer:
@spflueger The scope of the ADR has changed a bit as this is more mainly about packaging and a decision about version scheme. So I changed the title of the ADR and the PR and made a note that the automatisation will have to be figured out from practice.

redeboer:
Closed by https://github.com/ComPWA/expertsystem/pull/102

PR #71, opened 2020-06-19
feat: add  YAML adapter canonical amplitude 

redeboer:
Closes #71 

Only needed to add a converter for the CanonicalSum + Clebsch-Gordan sections, because `CanonicalAmplitudeGeneratorXML` inherits from the already tackled `HelicityAmplitudeGeneratorXML`.


expertsystem/amplitude/_yaml_adapter.py
```diff
@@ -320,6 +320,31 @@ def _extract_intensity_component(definition: Dict[str, Any]) -> Dict[str, Any]:
...
+                }
+                attributes = {
+                    key[1:]: value
+                    for key, value in clebsch_gordan_old.items()
+                    if key.startswith("@")
+                }
+                clebsch_gordan_new.update(attributes)
+                clebsch_gordan_dict[type_name] = clebsch_gordan_new
+            cano_sum_new["ClebschGordan"] = clebsch_gordan_dict
+            output_dict["CanonicalSum"] = cano_sum_new
```
spflueger, line 347: So I ran the test script and checked the output:
```yml
CanonicalSum:
    L: 2
    S: 1.0
    ClebschGordan:
        LS:
            J: 1.0
            M: -1.0
            j1: 2.0
            m1: 0.0
            j2: 1.0
            m2: -1.0
        s2s3:
            J: 1.0
            M: -1.0
            j1: 0.0
            m1: 0.0
            j2: 1.0
            m2: -1.0      
```
There is redundant information. I would propose this:

```yml
Canonical:
    LS:
        ClebschGordan:
            J: 1.0
            M: -1.0
            j1: 2.0
            m1: 0.0
            j2: 1.0
            m2: -1.0
    s2s3:
        ClebschGordan:
            J: 1.0
            M: -1.0
            j1: 0.0
            m1: 0.0
            j2: 1.0
            m2: -1.0      
```
I used the ClebschGordan label, because I want this to be like a mathematical command.

redeboer: Alright, I'll implement that, won't be hard. I trust that `L` and `S` are redundant here?

spflueger: Yeah, the j1 and j2 in the first block are L and S and couple to J. Also in the second block the J value is equal to S.

redeboer: ed2347f9494d933dbbbe1bdc66ad13a8e14dc270

expertsystem/amplitude/_yaml_adapter.py
```diff
@@ -320,6 +320,26 @@ def _extract_intensity_component(definition: Dict[str, Any]) -> Dict[str, Any]:
...
+                    "M": clebsch_gordan_old["M"],
+                }
+                attributes = {
+                    key[1:]: value
+                    for key, value in clebsch_gordan_old.items()
+                    if key.startswith("@")
+                }
+                clebsch_gordan_new.update(attributes)
+                cano_sum_new[type_name] = clebsch_gordan_new
+            output_dict["Canonical"] = cano_sum_new
```
spflueger, line 342: Hmm the `ClebschGordan` labels are missing

redeboer: Oops... well spotted

redeboer: 907acd9918f046dca17eb6121637112588dd4865

Issue #76, opened 2020-06-23
Equalize dev setup with tensorwaves


redeboer:
Now that https://github.com/ComPWA/tensorwaves/issues/78 has been successfully implemented, we can follow the same set-up for the expert system. For now, we only cannot check the entire framework with `mypy`, `flake8`, and `pylint` (e.g. #43).

A step plan (taken over from https://github.com/ComPWA/tensorwaves/issues/78):
- [x] Add VScode folder configuration
- [x] Improve [`pydocstyle`](https://pypi.org/project/pydocstyle/) configuration
- [x] Add [`rstcheck`](https://github.com/myint/rstcheck#configuration-file)
- [x] Add [`prettier`](https://prettier.io/) formatter for Markdown, as this appears to be the only tool that can hard-wrap lines through pre-commit
- [x] Enforce pre-commit through Travis CI on _all_ files (`pre-commit run -a`)
- [x] Explain the importance of linting tools in the contributing section (i.e., they formulate and automatise our convention policies)
- [x] Merge `pytest.ini` and `.coveragerc` with `tox.ini`
- [x] Create `tox` test environments
- [x] Fix `flake8` errors and remove conflicts with other linters
- [ ] Fix `pylint` errors and fix config problems for external packages
-  ~Fix `mypy` errors~ --> #43

PR #86, opened 2020-06-25
ci!: fix and enforce flake8 check

redeboer:
Part 7/8 of #76


expertsystem/amplitude/canonicaldecay.py
```diff
@@ -73,6 +76,73 @@ def generate_unique_amplitude_name(self, graph, node_id=None):
         return name
 
 
+def _clebsch_gordan_decorator(decay_generate_function):
```
spflueger, line 79: what is that move good for?

redeboer: `flake8` was complaining this is no method, but a function. First I changed it into a `staticmethod`, but then it's not a proper decorator anymore

PR #90, opened 2020-06-29
ci: add cspell support

redeboer:
Note that:
1. `cspell` has not been added as a pre-commit hook, because it requires the system to have `npm`
2. `cspell` is not enforced through Travis

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -13,5 +13,33 @@
     "ignorePaths": [
         "cspell.json"
     ],
-    "language": "en-US"
+    "language": "en-US",
+    "words": [
+        "docstrings",
+        "expertsystem",
+        "jupyter",
+        "mypy",
+        "numpy",
+        "permalinks",
+        "pydocstyle",
+        "pylint",
+        "pytest",
+        "repos",
+        "rstcheck",
+        "xmltodict"
+    ],
+    "ignoreWords": [
+        "commit's",
+        "envlist",
+        "esbenp",
+        "htmlcov",
+        "ipynb",
+        "nbsphinx",
+        "nbstripout",
+        "oneline",
+        "pylintrc",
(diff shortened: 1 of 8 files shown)


redeboer:
@spflueger An idea:
Enforce `cspell` through Travis with:
```bash
cspell $(git diff master --name-only)
```
Only problem is that this is only optimal for vscode users.

spflueger:
> @spflueger An idea:
> Enforce `cspell` through Travis with:
> 
> ```shell
> cspell $(git diff master --name-only)
> ```
> 
> Only problem is that this is only optimal for vscode users.

Hmm sounds ok to me. If someone would not be using vscode, then cspell could still be installed manually.

cspell.json
```diff
@@ -0,0 +1,45 @@
...
+        "esbenp",
+        "htmlcov",
+        "ipynb",
+        "nbsphinx",
+        "nbstripout",
+        "oneline",
+        "pylintrc",
+        "pyproject",
+        "rcfile",
+        "rtfd"
```
spflueger, line 43: how do you differentiate between the two categories? cspell says words would appear as suggestions, not really sure this works with vscode nicely

redeboer: https://github.com/ComPWA/expertsystem/pull/90/files#diff-6a35e6d99715b7e5c07424b1f7f9fea0R71-R75

redeboer: I tested it in vscode, the suggestions work well!

redeboer:
> Hmm sounds ok to me. If someone would not be using vscode, then cspell could still be installed manually.

Ah nvm, it seems it doesn't work so well: `npm` is needed to install `cspell` and that messes up the python virtual environments in Travis:
https://docs.travis-ci.com/user/languages/python/

PR #96, opened 2020-07-06
ci!: ignore mypy expertsystem

redeboer:
This is the first step to add type hints module-by-module (#43)

* `mypy` is activated over the `tests` folder so that you have checks on e.g. built-in types.
* `mypy` is ignored over the `expertsystem`, because type hints are (for now) completely lacking there. We can activate `mypy` module by module while improving the code base. This will also reintroduce `mypy` checks on the `expertsystem` under `tests`.
* Strings in the `__init__.py` files under `tests` have been removed as they are not required by `pydocstyle` anymore.
* *Important*: this new `mypy` configuration allows one to use `pre-commit` again without false-positives and therefore also `tox`.

_Follow-up to #95_


tests/__init__.py
```diff
@@ -1 +0,0 @@
-"""Required to ignore mypy errors in the tests folder."""
```
spflueger, line 1: why are they not needed anymore

redeboer: True, this is actually something that's debatable. They now just serve as a comment for why these files are there, but I thought it bothersome to have that line in each file `__init__.py` file under `tests`.

What do you think? Keep that line in or remove them from all those `__init__.py` files? And if not such a comment, explain in the documentation?

It's a pity this is just caused by the needs of `mypy`...

spflueger: Why we need those `__init__.py` in the first place. This flag won't help https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-namespace-packages?

redeboer: Good idea! I just tried setting `namespace_packages = True` in the `tox.ini:mypy` section. However, `namespace_packages` makes `mypy` ignores the option
```ini
[mypy-tests.*]
disallow_untyped_defs = False
```
That's why the text:
```python
"""Required to ignore mypy errors in the tests folder."""
```
made sense, actually. So I'll put those docstrings back.

PR #97, opened 2020-07-06
ci: improve tox usage with more specific jobs

redeboer:
* Speed up main `tox` job by avoiding check on external links (`linkcheck`) and not executing Jupyter notebooks
* Define `docnb` job for executing notebooks (and building them into the docs)
* Simplify command `tox -e sty` (previously `tox -e style`)

_Follow-up to #96_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ jobs:
         - npm install -g cspell
       script:
         - pip3 install -r requirements-dev.txt
-        - tox -e style
+        - tox -e sty
         - cspell $(git diff d4f1928 --name-only)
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         - npm install -g cspell
       script:
         - pip3 install -r requirements-dev.txt
-        - tox -e style
```
spflueger, line 40: Just a comment. I personally would have preferred style because its easier to read

redeboer: Ah yeah that's right. I just tried to mimic what `tox` does, like `py` for the main test environment

PR #99, opened 2020-07-06
refactor: add XML root tag

### LINKED ISSUE #77: Add root element to XML output recipe file
XML files that the expert system produces are formally invalid, because they miss an [XML root element](
https://en.wikipedia.org/wiki/Root_element). This can for instance be seen when you run the [`check-xml`](https://github.com/pre-commit/pre-commit-hooks#check-xml) of pre-commit:

```bash
pre-commit run check-xml -a
```

See also
https://github.com/ComPWA/expertsystem/blob/d4f1928ccaaa208f36f428d5dfdb8a94fa75fbbc/tox.ini#L42-L44

In addition, if there is no single (!) root tag, `xmltodict` cannot `unparse` the file.

redeboer:
Closes #77
The lack of a root tag made it impossible to *import* an XML recipe file with `xmltodict`.

Also adds a VSCode debug file that makes it easier to debug individual Python files.

_Follow-up to #98_


particle_list.xml
```diff
@@ -1,3695 +1,3623 @@
...
-      <Type>Strangeness</Type>
-      <Value>0</Value>
-    </QuantumNumber>
-    <QuantumNumber>
-      <Class>Int</Class>
-      <Type>BaryonNumber</Type>
-      <Value>0</Value>
-    </QuantumNumber>
-  </Particle>
-
```
redeboer, line 142: Let me add back these, they were removed by mistake

examples/python/D0ToKs0KpKm.xml
```diff
@@ -1,417 +1,418 @@
 <?xml version="1.0" encoding="utf-8"?>
-
```
spflueger, line 2: why are these two xml files here in the first place. These are left overs from a test verifying the angular distribution

expertsystem/state/particle.py
```diff
@@ -305,19 +305,21 @@ def to_dict(input_ordered_dict: OrderedDict) -> dict:
     name_label = Labels.Name.name
     with open(file_path, "rb") as xmlfile:
         full_dict = xmltodict.parse(xmlfile)
-        for particle_definition in full_dict["ParticleList"]["Particle"]:
-            particle_name = particle_definition[name_label]
-            particle_list[particle_name] = to_dict(particle_definition)
+        full_dict = full_dict.get("root", full_dict)
+    for particle_definition in full_dict["ParticleList"]["Particle"]:
+        particle_name = particle_definition[name_label]
+        particle_list[particle_name] = to_dict(particle_definition)
```
spflueger, line 311: shouldn't this be indented to be inside the with scope?

examples/python/D0ToKs0KpKm.xml
redeboer: True. But I don't want to address those (indeed malfunctioning) examples in this PR. An issue can be made for this

expertsystem/state/particle.py
redeboer: Why?

examples/python/D0ToKs0KpKm.xml
spflueger: but cant this be just deleted instead of reformatting the whole file and an deleting it afterwards again?

spflueger: if you want this pr to be clean then I would suggest to apply a separate chore PR before that

expertsystem/state/particle.py
spflueger: Oh i thought with would create a scope, but it doesn't. Nvm then

examples/python/D0ToKs0KpKm.xml
redeboer: In that case we'd better completely remove those examples

redeboer: and I'm not sure if we really want that, that's why I say, better make an issue for this

redeboer: Will be rebased onto #101

Issue #23, opened 2020-05-13
Get particle info from PDG


redeboer:
- [x] Investigate how to get all info required to construct a `Particle` instance from the scikit-hep [`Particle`](https://pypi.org/project/Particle/) package. This would be _general particle info_:
  https://github.com/ComPWA/expertsystem/blob/8b5c257c6738d3e50ff1d56a7a06714e3e2a7036/expertsystem/data.py#L176-L183
  and _quantum numbers_:
  https://github.com/ComPWA/expertsystem/blob/8b5c257c6738d3e50ff1d56a7a06714e3e2a7036/expertsystem/data.py#L113-L126

- [x] Loop over the PDG definitions and create `Particle` instances, then add them with `DATABASE.add`. This would have to be done under `load_default_particle_list`, right here:
  https://github.com/ComPWA/expertsystem/blob/0265e9dc3b8f6cc5ae7c2d9ac834fdc28335e5ba/expertsystem/ui/__init__.py#L376
  See here for a syntax example:
  https://github.com/ComPWA/expertsystem/blob/8b5c257c6738d3e50ff1d56a7a06714e3e2a7036/tests/io/test_particle_collection.py#L21-L33
  Note: we use the template `[float]` instead of `[Spin]` because we are only interested in spin _magnitude_, not projection.

- [ ] Check whether definitions in `particle_list.yml` agree with the ones imported from the PDG.

redeboer:
@sebastianJaeger @Leongrim 
I added some snippets to the issue description :) Let me know if it you have questions!

Leongrim:
I looked into the `particle` package from `scikit-hep`. There are a few problems with it.

This package does not get the information from the webpage but from a `.csv` file which comes with the package which does not include all the particles that are found on the webpage.

The larger issue is not all quantum numbers are accessible. Something like Bottomness or IsospinZ are still calculable from the quarks. But regarding the lepton numbers there seems to be no a work around.

redeboer:
> This package does not get the information from the webpage but from a .csv file which comes with the package which does not include all the particles that are found on the webpage.

Yes, this is true. But I am not aware of an official API for the PDG, so I think this package is the closest we can get, that is, 'outsource the downloading' to that package.

> The larger issue is not all quantum numbers are accessible. Something like Bottomness or IsospinZ are still calculable from the quarks. But regarding the lepton numbers there seems to be no a work around.

[...]

PR #100, opened 2020-07-06
feat!: add ParticleCollection structures

redeboer:
* Validate `particle_list.yml` with [JSON Schema](https://json-schema.org/)
* Load and dump `particle_list.yml` from and to a `ParticleCollection` class
* `ParticleCollection` class consists of `Particle` instances. These will be converted into nested `dict`s for internal consumption in a follow-up PR

_Follow-up to #99_


redeboer:
Here we go again... https://github.com/ComPWA/tensorwaves/issues/28 needs to be tackled asap. Because now it's the schema files that are not installed... (see e.g. https://readthedocs.org/projects/expertsystem/builds/11399166/)

redeboer:
Problems mentioned in https://github.com/ComPWA/expertsystem/pull/100#issuecomment-654332914 have been fixed by #102. This PR has been rebased on that fix.

expertsystem/data.py
```diff
@@ -0,0 +1,191 @@
...
+        return self.__max
+
+    @max_value.setter
+    def max_value(self, value: Optional[float]) -> None:
+        if value is not None and value < self.value:
+            raise ValueError(
+                "Cannot set maximal value lower than value"
+                f" ({value} < {self.value})"
+            )
+        self.__max = value
```
spflueger, line 65: Is this used anywhere in the code? If not then I would suggest for this to be removed again, until the point it will be used.

expertsystem/data.py
```diff
@@ -0,0 +1,191 @@
...
+    def __init__(self, magnitude: float, projection: float) -> None:
+        if abs(projection) > magnitude:
+            raise ValueError(
+                "Spin projection cannot be larger than its magnitude"
+            )
+        self.__magnitude = float(magnitude)
+        self.__projection = float(projection)
+
+    def __eq__(self, other: object) -> bool:
+        return self.__magnitude == other
```
spflueger, line 105: Here I would do a full comparison. A magnitude comparison should be done via `self.magnitude == other.magnitude`

expertsystem/data.py
```diff
@@ -0,0 +1,191 @@
...
+    def __eq__(self, other: object) -> bool:
+        return self.value == other
+
+    def __float__(self) -> float:
+        return self.value
+
+    def __repr__(self) -> str:
+        if self.uncertainty is None:
+            return str(self.value)
+        return f"{self.value} ± {self.uncertainty}"
```
spflueger, line 137: In my opinion this is a Parameter, so a pair of a Value and an Error

expertsystem/data.py
```diff
@@ -0,0 +1,191 @@
...
+    spin: float
+    mass: MeasuredValue
+    strange: int = 0
+    charm: int = 0
+    bottom: int = 0
+    top: int = 0
+    baryon: int = 0
+    ln_e: int = 0
+    ln_mu: int = 0
+    ln_tau: int = 0
```
spflueger, line 155: We should think about the names. I would tend to use the official names, even if they are slightly longer

expertsystem/data.py
```diff
@@ -0,0 +1,191 @@
...
+        self.__particles[particle.name] = particle
+
+    def items(self) -> ItemsView[str, Particle]:
+        return self.__particles.items()
+
+    def keys(self) -> KeysView[str]:
+        return self.__particles.keys()
+
+    def values(self) -> ValuesView[Particle]:
+        return self.__particles.values()
```
spflueger, line 191: I think this should inherit from collections base class, as the behavior is something like a dict but not completely. Which could be confusing to a user.

expertsystem/io/__init__.py
```diff
@@ -0,0 +1,39 @@
+"""Serialization module for containers of `expertsystem.data`."""
```
spflueger, line 1: Just to make sure I understand this. Later on in the refactored expertsystem, this whole `io` package can be reduced to merely the load and dump functions, right?

expertsystem/particle_list.yml
```diff
@@ -2,16 +2,13 @@
...
     QuantumNumbers:
       Spin: 1
       Charge: 0
       Parity: -1
       Cparity: -1
-      IsoSpin: { Value: 0, Projection: 0 }
-      BaryonNumber: 0
-      Charm: 0
-      Strangeness: 0
+      IsoSpin: 0
```
spflueger, line 11: Is the code able to set the projection to zero in this case as well?

expertsystem/schemas/yaml-particle-list.json
```diff
@@ -0,0 +1,85 @@
...
+        "Charmness": { "type": "integer" },
+        "Bottomness": { "type": "integer" },
+        "Topness": { "type": "integer" },
+        "BaryonNumber": { "type": "integer" },
+        "ElectronLN": { "type": "integer" },
+        "MuonLN": { "type": "integer" },
+        "TauLN": { "type": "integer" },
+        "IsoSpin": {
+          "anyOf": [
+            { "type": "number", "minimum": 0, "maximum": 0 },
```
spflueger, line 46: What is this maximum about?

tests/data/test_parameter.py
```diff
@@ -0,0 +1,28 @@
+import pytest
```
spflueger, line 1: I would postpone to add this Parameter code until it will actually be used

expertsystem/data.py
redeboer: Yep... good point

spflueger:
Another comment: This `ParticleCollection` is not used in the expertsystem yet or? Would be good to have some connection to the expert system core, otherwise its just some standalone/unused codebase for now. Or is it the plan to attach it in a follow up PR?

expertsystem/data.py
redeboer: Good one. Actually what this thing is for is to allow for instance:
```python
assert isospin == 0.5
```
So comparison with a `float` or `int`.

expertsystem/data.py
redeboer: One may wonder what that error is doing here at all... In the expertsystem, only the value is used (namely mass and width)

expertsystem/data.py
redeboer: You mean for instance `lepton_number_e`?

expertsystem/data.py
redeboer: Actually I was searching for a `dict` base class, but there is no such `abc` as far as I could find, also not in the `typing` library. And technically, this `ParticleCollection` is like a 'read-only `dict`'.

expertsystem/particle_list.yml
redeboer: Yep, there's a check on this. In fact, it will even complain if:
* `abs(projection) > magnitude`
* `magnitude` > 0 and no projection is provided

expertsystem/schemas/yaml-particle-list.json
redeboer: https://github.com/ComPWA/expertsystem/pull/100#discussion_r450957428

tests/data/test_parameter.py
redeboer: Ok

expertsystem/data.py
spflueger: So one example I'm referring to is the `__iter__` method. I guess this should be a `Mapping` + the add function

expertsystem/io/__init__.py
redeboer: Yes I think this `io` module will eventually have only two, maybe three, wrapper functions:
* `write`
* `load`

I say three, because the `load` would have to take its interpretation from the recipe file. Then it would produce a `ParticleCollection` instance if the recipe file only contains a `"ParticleList"` section and a `Recipe` instance if the recipe file contains more than that. But to be honest, I think that's a bit to much implicit reasoning, so I would prefer a destinction here: `load_particle_collection` (as introduced in this PR) and `load_recipe` (follow-up PRs).

Note that `write` can get all the information from its input with `isinstance` checks, so one `write` function is sufficient (it takes it's output format from the file extension.

However, even on the surface, this module only offers two/three functions, there are a lot more implementation details underneath, so I would still keep this `io` module separate.

expertsystem/data.py
spflueger: yeah, `strangeness`, `baryon_number` etc

expertsystem/data.py
redeboer: Yes, this is a good point actually... The missing `__iter__` method actually causes a subtle bug in the tests, which, clumsily, had been avoided by iterating over `.keys()`.

redeboer: So you say we would need to define our own abstract base class? I cannot think of a built-in abc that has this kind of read-only interface.

expertsystem/data.py
redeboer: Ah ok. Yes for `strange` I had in mind something like `j_psi.strange = 0`. But I guess `j_psi.strangeness = 0` is more natural indeed... Only the `lepton_number_e`... meh, not sure if I would prefer that over `ln_e`.

expertsystem/data.py
spflueger: No, just use the matching collection abc. So `class ParticleCollection(abc.Mapping):`. Then we ensure it behaves just like a Mapping. In addition we add this `add` method (lots of adds :smile: ).

spflueger: Maybe this also suggests that we are using a wrong thing. Theoretically you don't need the keys and you just have a hash set based on the pid or name.

redeboer: Yeah sorry I just realised when looking at the `collections.abc` class... :see_no_evil:

redeboer: > Maybe this also suggests that we are using a wrong thing. Theoretically you don't need the keys and you just have a hash set based on the pid or name.

I don't think so. One needs some sort of lookup, right?

spflueger: True

redeboer:
> Another comment: This `ParticleCollection` is not used in the expertsystem yet or? Would be good to have some connection to the expert system core, otherwise its just some standalone/unused codebase for now. Or is it the plan to attach it in a follow up PR?

Yes, it's used in the next PR (on XML-to-`ParticleCollection`). But if you want, I can already implement `ParticleCollection` in the existing `_yaml_adapter` (that is currently lined up for a PR _after_ XML-to-ParticleCollection)

redeboer:
> But if you want, I can already implement `ParticleCollection` in the existing `_yaml_adapter` (that is currently lined up for a PR after XML-to-ParticleCollection)

Sorry, that's incorrect. Before implementing `ParticleCollection` in `_yaml_adapter`, I need a dumper and loader to XML (that is, internal `dict` structure). That will be added in the next PR.

spflueger:
> > Another comment: This `ParticleCollection` is not used in the expertsystem yet or? Would be good to have some connection to the expert system core, otherwise its just some standalone/unused codebase for now. Or is it the plan to attach it in a follow up PR?
> 
> Yes, it's used in the next PR (on XML-to-`ParticleCollection`). But if you want, I can already implement `ParticleCollection` in the existing `_yaml_adapter` (that is currently lined up for a PR _after_ XML-to-ParticleCollection)

So that step will replace the particle DB? I just would be good to see this ParticleCollection in action to judge the code.

redeboer:
> I just would be good to see this ParticleCollection in action to judge the code.

Can do all in one, I guess, but the PR would be massive...

> So that step will replace the particle DB?

Two things can happen once the XML-to-`ParticleDatabase` (next PR) is through:
1. `_yaml_adapter` (which is a bit of a hack for now anyway as long as there is no `Recipe` class like `ParticleDatabase`) can be simplified by using the `io` module.
2. use `ParticleDatabase` as starting point. This won't yet replace the `particle.DATABASE` (because we need those nested `dict`s for now) but it will allow us to switch source of truth easily (e.g. #23). It's a bit hard to explain, but see diagram below.

ExpertSystem-TensorWaves integration

PR #101, opened 2020-07-06
chore: remove Python examples

redeboer:
See https://github.com/ComPWA/expertsystem/pull/99#discussion_r450211976


examples/python/epem_to_chicpipi.py
```diff
@@ -1,37 +0,0 @@
-r"""Sample script for the testing purposes.
```
spflueger, line 1: why delete the python scripts?

redeboer: Oh thought those were the ones that were outdated? For sure they are not under CI

redeboer: Okay seems they do work actually. So, shall I put them back? And, if so, should a Travis job be made that tests them?

spflueger: I guess they are not in CI because they are slow? Normally this would still run though or serve as some kind of example. The xml files are recipes for tensorwaves, so they def do not belong here.

spflueger: Hmm i think the tricky part is the speed. They are just too slow for CI. Not sure if they have any value as an example for a user...

redeboer: Yeah, the notebooks already serve the purpose of documentation

redeboer: If we do remove the Python scripts, I should also move `examples/jupyter` --> `examples`

redeboer: Actually it's only `epem_to_d0d0barpipi.py` that takes ages

spflueger: all right, lets remove them then

PR #102, opened 2020-07-06
ci!: single source of truth for dependencies 🎉

### LINKED ISSUE #103: Improve requirements structure
- [x] Single source of truth for requirements (`setup.cfg` or `pyproject.toml`)
- [x] Improve developer installation (e.g. `pip install -e .`)
  (currently mainly through `environment.yml`, but that causes problems, see https://github.com/ComPWA/tensorwaves/issues/53#issuecomment-651008386)
- [x] Test, documentation, and dev requirements through same source of truth?
- ~Extract package version from latest Git tag~ --> #68 

_Original description:_
> For me the general way the requirements are currently setup is not ideal.
> https://stackoverflow.com/questions/17803829/how-to-customize-a-requirements-txt-for-multiple-environments
> I would follow this. So having requirements files for each environment
> - ci/test
> - prod
> - dev

> and reuse common dependencies across them. Then if you are a dev, you simply run the dev req and you are all setup and so on.

redeboer:
Closes #103 

* Contain all requirements in `setup.cfg`
* This simplifies the Travis config, install instructions, and contribute instructions
* Also solves the issues of installing package data (e.g. https://github.com/ComPWA/expertsystem/pull/100#issuecomment-654332914)

Remaining issues:
- [x] Fix [RTD config file](https://docs.readthedocs.io/en/stable/config-file/v2.html), see [failed job](https://readthedocs.org/projects/expertsystem/builds/11401038/)
- [ ] **Question** Possibly should not freeze the package requirements itself, see https://github.com/ComPWA/expertsystem/pull/102#issuecomment-654389180
- ~Version should be extracted from Git tag at deploy stage, not `setup.cfg`.~ --> issue #68


redeboer:
@spflueger with regard to 558b850df61d9306c3f8478740377fd8ca596578:
While freezing the requirements is something we wanted to do for some time now, it is super unfriendly for other packages that would list the expertsystem as a dependency, because strict dependencies have a high chance to result in dependency conflicts for the caller higher up (read: `tensorwaves`).

One could say only the `requirements-dev.txt` may therefore be frozen -- the actual package dependencies should be kept as loose as possible.

spflueger:
> @spflueger with regard to 558b850df61d9306c3f8478740377fd8ca596578:
> While freezing the requirements is something we wanted to do for some time now, it is super unfriendly for other packages that would list the expertsystem as a dependency, because strict dependencies have a high chance to result in dependency conflicts for the caller higher up (read: `tensorwaves`).
> 
> One could say only the `requirements-dev.txt` may therefore be frozen -- the actual package dependencies should be kept as loose as possible.

I think if you have a dependency conflict then you just need to solve that. I would say for the combination of tensorwaves and the expertsystem it should be rather easy to handle that. The pinning will just help us to ensure we do not experience incompatibilities on different systems and reproducibility

redeboer:
> I think if you have a dependency conflict then you just need to solve that.

Sure, if the `expertsystem` is only used in `tensorwaves`, then this is fine. But that's why I say: freezing isn't a great service. Because if your package depends on the `expertsystem`, you are forced to follow its dependencies precisely (and may run into problems when you depend on package that have other specific requirements.

redeboer:
@spflueger Here we go already:
https://travis-ci.com/github/ComPWA/expertsystem/jobs/358244484#L497
Although in this case it's actually a conflict in the dev requirements, hadn't expected that

redeboer:
d7f420301f8dd0a6e5ba1b32136414629db12e99 introduces a bug. While `tests` is indeed not installed anymore, all subpackages of `expertsystem` are not installed either...

setup.cfg
```diff
@@ -0,0 +1,81 @@
...
+    Programming Language :: Python
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Topic :: Scientific/Engineering
+    Topic :: Scientific/Engineering :: Physics
+    Typing :: Typed
+
+[options]
+python_requires = >=3.6
```
spflueger, line 33: should we also add some pip requirement?

redeboer: I guess technically we should (not sure how) but then we need to know more about what we expect of pip

spflueger: I can't think of any problems so far and it seems to work with pip 9.0.1. We can add it if we experience any problems

PR #110, opened 2020-07-09
refactor!: ParticleCollection -> particle.DATABASE

redeboer:
_Breaking change: decay info is not added to the recipe anymore! Therefore the tests had to be modified and functionality has been lost. Dynamics has to be added by the user in another phase, but this requires a big change in the interface_

Note also that the values of the internal nested `dict`s of `particle.DATABASE` have become `int`/`float`, not `str`. That's why the `QNConverter` converter classes don't cast to `str` anymore (see also #122)

### DIFF
--- a/tests/amplitude/test_yaml_helicity.py
+++ b/tests/amplitude/test_yaml_helicity.py
@@ -70,7 +70,9 @@ class TestHelicityAmplitudeGeneratorYAML:
         assert len(recipe) == 3
 
     def test_particle_section(self):
-        particle_list = self.imported_dict["ParticleList"]
+        particle_list = self.imported_dict.get(
+            "ParticleList", self.imported_dict
+        )
         gamma = particle_list["gamma"]
         assert gamma["PID"] == 22
         assert gamma["Mass"] == 0.0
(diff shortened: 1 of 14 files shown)


spflueger:
> _Breaking change: decay info is not added to the recipe anymore! Therefore the tests had to be modified and functionality has been lost. Dynamics has to be added by the user in another phase, but this requires a big change in the interface_

So what is the plan of inserting the dynamics into the model? We need some plan on how we want this to look like at the end and how things work together

redeboer:
> So what is the plan of inserting the dynamics into the model? We need some plan on how we want this to look like at the end and how things work together

Yep... but this is a bigger issue. For some reason, the expertsystem currently assumes you provide a particle `dict` **with** dynamics (and dynamics should not part of a particle instance imho). So we have to rethink our interface... add some place where they user can provide, say, a mapping like (just a random example):

```python
added_dynamics = {
    "f0(980)": "relativisticBreitWigner",
    "f0(1500)": "nonResonant",
}
```

Question is, where?

tests/amplitude/test_yaml_helicity.py
```diff
@@ -105,22 +107,22 @@ def test_parameter_section(self):
 
     def test_dynamics_section(self):
         dynamics = self.imported_dict["Dynamics"]
-        assert len(dynamics) == 5
+        assert len(dynamics) == 1
```
spflueger, line 110: This change is correct? I did not verify in the recipe. Just want to make sure this is supposed to change

redeboer: This is because this PR does the following:
`particle_list.xml` --> `DATABASE` [nested dict]
becomes
`particle_list.xml` -> `ParticleCollection`  --> `DATABASE`  [nested `dict`]
Since dynamics are not retained in the `ParticleCollection`, this information is lost, so the dynamics section of the recipe file shrinks.

spflueger: So how is the the dynamics added to the final recipe yml then?

redeboer: https://github.com/ComPWA/expertsystem/pull/110#issuecomment-656055278

spflueger: Seems like i need to dig a bit deeper into these code changes

spflueger: But then i expect this thing to be empty? But somehow the `J/Psi` in this example is in there

redeboer: Yeah because initial state is always given non-dynamic. Not sure where, I can look it up

spflueger: That is added in the helicity amplitude generator. Then we should also remove this part

spflueger: all right as it triggers another bigger change lets move this to a follow up pr

PR #113, opened 2020-07-09
build!: implement setuptools-scm

### LINKED ISSUE #68: Standardize and automate release flow

redeboer:
Closes #68

### DIFF
--- a/doc/contribute.rst
+++ b/doc/contribute.rst
@@ -18,6 +18,7 @@ source code is considered the 'installation'.
     git clone git@github.com:ComPWA/tensorwaves.git
     code tensorwaves
     conda env create
+    conda activate es
     pip install -e .[dev]
 
   For more info, see :ref:`contribute:Visual Studio code`.
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,5 @@
 [metadata]
 name = expertsystem
-version = 0.1.2-alpha
 author = The ComPWA Team
 author_email = compwa-admin@ep1.rub.de
 maintainer_email = compwa-admin@ep1.rub.de
@@ -31,6 +30,8 @@ classifiers =
 
 [options]
 python_requires = >=3.6
+setup_requires =
+    setuptools_scm>=30.3.0
 install_requires =
     jsonschema==3.2.0
     numpy==1.19.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ classifiers =
 [options]
 python_requires = >=3.6
 setup_requires =
-    setuptools_scm>=30.3.0
+    setuptools_scm
 install_requires =
     jsonschema==3.2.0
     numpy==1.19.0
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -10,8 +10,8 @@ assignees: ""
 concise description of what the problem is. [Example: "I'm always frustrated
 when ..."]
 
-**Describe the solution you'd like** A clear and concise description of what you
-want to happen.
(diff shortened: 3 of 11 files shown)


setup.cfg
```diff
@@ -31,6 +30,8 @@ classifiers =
 
 [options]
 python_requires = >=3.6
+setup_requires =
+    setuptools_scm>=30.3.0
```
spflueger: whats the difference between this one and above

doc/contribute.rst
```diff
@@ -18,6 +18,7 @@ source code is considered the 'installation'.
     git clone git@github.com:ComPWA/tensorwaves.git
     code tensorwaves
     conda env create
```
spflueger, line 20: actually dont you have to specify some name of the venv here?

redeboer: No, the name is specified in the `environment.yml` file. And you also don't have to use the flag `--file` because `environment.yml` is the default for that flag when running `conda env create`.

setup.cfg
redeboer: What do you mean by above?

spflueger: in the pyproject.toml

doc/contribute.rst
spflueger: ah i see

setup.cfg
spflueger: and why is the version not pinned?

redeboer: Yeah emm.. good point. You can read up on the documentation [here](https://pypi.org/project/setuptools-scm/). I'm not 100% sure, I just added this statement here as well, assuming `pip` or `setup.py` uses `setup.cfg`, not `pyproject` (while `setuptools-scm` merely uses `pyproject.toml` as a config).

redeboer: > and why is the version not pinned?

Oh but this I certainly wouldn't do... That would mean that you require the install system to have one specific version. Almost like requiring one specific version of python.

spflueger: Hmm, but we are not fixing Python just all the dependencies right? I just want to avoid any troubles with incompatibilities

redeboer: I think we only want to fix dependencies that come during the install stage, that is, the ones listed under `install_requires`.

PR #114, opened 2020-07-20
ci: fix setuptools-scm deploy

redeboer:


.travis.yml
```diff
@@ -54,15 +54,14 @@ jobs:
...
-      install: echo "Skipping install stage"
+      install:
+        - pip3 install .
       script: echo "Skipping script stage"
       deploy:
         edge: true
         provider: pypi
         on:
           tags: true
-        distributions: sdist
```
redeboer, line 64: Removed these because afaics we should the travis defaults should be fine for `setuptools-scm`

.travis.yml
```diff
@@ -54,15 +54,14 @@ jobs:
       language: shell
 
     - name: "Deploy PyPI package"
-      install: echo "Skipping install stage"
+      install:
```
redeboer, line 57: I don't think installing is necessary, but at least I'm sure this creates a version file. Not sure what the `deploy` stage with `provider: pypi` does exactly.

.travis.yml
```diff
@@ -54,15 +54,14 @@ jobs:
...
+        - pip3 install .
       script: echo "Skipping script stage"
       deploy:
         edge: true
         provider: pypi
         on:
           tags: true
-        distributions: sdist
-        skip_existing: false
```
spflueger, line 65: Why are those removed?

.travis.yml
spflueger: Where does it create the version file in the install location?

.travis.yml
spflueger: Hmm i see. Maybe its better to keep the `skip_existing` in case they change the default value of that

.travis.yml
redeboer: https://github.com/ComPWA/expertsystem/blob/4fc2df6c21dd0c9fe611144b1de8d01b77a238c0/pyproject.toml#L9-L10

spflueger: can't you test that the version will be correct locally. In addition you can also use https://test.pypi.org/ to upload

redeboer: Thing is that I'm not sure what this `deploy` job does exactly on Travis.

spflueger: well build the sdist and upload it to pypi

redeboer: > This feature requires Setuptools 42 or later, released in Nov, 2019. **If your project needs to support build from sdist on older versions of Setuptools, you will need to also implement the `setup.py usage` for those legacy environments.**

Not sure what Travis does it you remove `sdist`

redeboer: > well build the sdist and upload it to pypi

but yeah `sdist` does not work, also locally, it does not produce the version file

spflueger: So does the current code work? I want to avoid doing trail and error here

redeboer: No, also not `bdist_wheel`

redeboer: Should work now :crossed_fingers:

PR #121, opened 2020-07-23
feat: add write amplitude model from STM

### LINKED ISSUE #91: HelicityAmplitudeGenerator should warn if not using the correct formalism
Currently, it is possible to use the expert system with, say, the `StateTransitionManager` with argument `formalism_type='canonical-helicity'` in combination with `HelicityAmplitudeGeneratorXML` and vice versa. An exception should be raised in this situation, because the combination physically doesn't make sense.

redeboer:
Closes #91 

Amplitude model recipe writing can now be done directly from the STM. This removes the need to first create e.g. a `HelicityAmplitudeGenerator`.

### DIFF
--- a/expertsystem/ui/__init__.py
+++ b/expertsystem/ui/__init__.py
@@ -165,99 +165,6 @@ class StateTransitionManager:  # pylint: disable=too-many-instance-attributes
         )
         return graph_settings_groups
 
-    def find_solutions(
-        self, graph_setting_groups
-    ):  # pylint: disable=too-many-locals
-        """Check for solutions for a specific set of interaction settings."""
-        results = {}
-        logging.info(
-            "Number of interaction settings groups being processed: %d",
-            len(graph_setting_groups),
-        )
-        for strength, graph_setting_group in sorted(
-            graph_setting_groups.items(), reverse=True
-        ):
-            logging.info(
-                "processing interaction settings group with "
-                f"strength {strength}",
-            )
-            logging.info(f"{graph_setting_group} entries in this group")
-            logging.info(f"running with {self.number_of_threads} threads...")
-
-            temp_results = []
-            progress_bar = IncrementalBar(
-                "Propagating quantum numbers...", max=len(graph_setting_group)
-            )
-            progress_bar.update()
-            if self.number_of_threads > 1:
-                with Pool(self.number_of_threads) as pool:
-                    for result in pool.imap_unordered(
-                        self._propagate_quantum_numbers, graph_setting_group, 1
-                    ):
-                        temp_results.append(result)
-                        progress_bar.next()
-            else:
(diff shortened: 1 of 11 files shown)


expertsystem/ui/__init__.py
```diff
@@ -239,84 +339,6 @@ def _determine_node_settings(self, graphs):
             graph_node_setting_pairs.append((instance, node_settings))
         return graph_node_setting_pairs
 
-    def find_solutions(
```
spflueger, line 242: Why did you move this up?  My logic here was to keep functions called by other functions close to each other, so that the user does not need to scroll around too much

redeboer: To keep the public functions together (1c7216aec94b0370b8f58ae7113fe455debc6394). It's more for the developer, in that sense it's indeed better to keep it on the same place

expertsystem/ui/__init__.py
```diff
@@ -72,6 +74,7 @@ def __init__(  # pylint: disable=too-many-arguments
                 f"Formalism type {formalism_type} not implemented."
                 f" Use {allowed_formalism_types} instead."
             )
+        self.__formalism_type = formalism_type
```
spflueger, line 77: Actually a lot of the other attributes should also be made into private variables or properties etc

redeboer: Yep... working on that

redeboer: but better to do that in a separate PR

expertsystem/ui/__init__.py
redeboer: 18e8684d48fcb993e71451578fd688a335873009

expertsystem/ui/__init__.py
spflueger: ok

expertsystem/ui/__init__.py
spflueger: Ok. Although I would be fine with such a change in general. We should have some kind of decision on this and then write this down in and adr or the docs. I want to avoid such changes going back and forth and having no consistent manner.

redeboer: It's hard to pin down. Sometimes you need to have the class defined first. Maybe one prefers public methods first. And then in some cases having the definition coming right after the call is easier for readability.

spflueger: I just followed the recommendation from clean code i believe (which made sense to me). I guess nowadays with strong IDEs this not as crucial anymore, but i think its still kind of nice to have thinks stuck together. So I would actually vote for how it is now. But I can be convinced otherwise :)

redeboer: Yep it's from Clean Code ^^

PR #135, opened 2020-07-29
feat: define quantum state interfaces

### LINKED ISSUE #127: Extract quantum state related classes from `Particle`
Currently, a `Particle` is just a source of truth, while in the eventual graph, we need to work with quantum numbers only (that is, we are not interested in PID and name).

The aim (https://github.com/ComPWA/expertsystem/issues/127#issuecomment-665717889):

- [x] `Particle` - statically defines particle info from in the PDG [immutable]
- [ ] `QuantumState` - dynamically defines a set of quantum numbers (used in `graph`, `solvers`, and `conservation_rules`) [immutable or mutable]
- [ ] `ComplexEnergyState` - statically defines a fully defined particle state [immutable]
  Used for 'mass and width' (that is, pole position) and therefore eventually ends up in the amplitude model.

redeboer:
Closes #127

**Major changes**
* Dropped support for Python 3.6 in order to use [`dataclasses`](https://docs.python.org/3/library/dataclasses.html)
* Removed type hints, see #149

### DIFF
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -43,16 +43,19 @@ class Parity:
 class Spin(abc.Hashable):
     """Safe, immutable data container for (iso)spin."""
 
-    def __init__(self, magnitude: float, projection: float) -> None:
+    def __init__(
+        self, magnitude: float, projection: Optional[float] = None
+    ) -> None:
         magnitude = float(magnitude)
-        projection = float(projection)
-        if abs(projection) > magnitude:
-            raise ValueError(
-                "Spin projection cannot be larger than its magnitude:\n"
-                f"  {projection} > {magnitude}"
-            )
-        if projection == -0.0:
-            projection = 0.0
+        if projection is not None:
+            projection = float(projection)
+            if abs(projection) > magnitude:
+                raise ValueError(
+                    "Spin projection cannot be larger than its magnitude:\n"
+                    f"  {projection} > {magnitude}"
+                )
+            if projection == -0.0:
+                projection = 0.0
         self.__magnitude = magnitude
         self.__projection = projection
 
@@ -75,7 +78,7 @@ class Spin(abc.Hashable):
         return self.__magnitude
 
     @property
-    def projection(self) -> float:
+    def projection(self) -> Optional[float]:
         return self.__projection
 
     def __hash__(self) -> int:
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -117,7 +117,10 @@ class ComplexEnergyState:
 
     def __eq__(self, other: object) -> bool:
         if isinstance(other, ComplexEnergyState):
-            return self == other
+            return (
+                self.complex_energy == other.complex_energy
+                and self.state == other.state
+            )
         raise NotImplementedError
 
     def __repr__(self) -> str:
@@ -163,8 +166,6 @@ class Particle(ComplexEnergyState):
                 and self.complex_energy == other.complex_energy
                 and self.state == other.state
             )
-        if isinstance(other, ComplexEnergyState):
-            return super() == other
         raise NotImplementedError
 
     def __repr__(self) -> str:
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -108,37 +108,83 @@ class QuantumState(NamedTuple):
     g_parity: Optional[Parity] = None
 
 
-class ComplexEnergyState(NamedTuple):
+class ComplexEnergyState:
     """Pole in the complex energy plane, with quantum numbers."""
 
-    energy: complex
-    state: QuantumState
+    def __init__(self, energy: complex, state: QuantumState):
+        self.__energy = complex(energy)
+        self.state: QuantumState = state
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, ComplexEnergyState):
+            return self == other
+        if isinstance(other, QuantumState):
+            return self.state == other
+        raise NotImplementedError
+
+    def __repr__(self) -> str:
+        return str(self.__energy)
+
+    @property
+    def complex_energy(self) -> complex:
+        return self.__energy
+
+    @property
+    def mass(self) -> float:
+        return self.__energy.real
+
+    @property
+    def width(self) -> float:
+        return self.__energy.imag
 
 
-class Particle(NamedTuple):
+class Particle(ComplexEnergyState):
     """Immutable container of data defining a physical particle.
 
     Can **only** contain info that the `PDG <http://pdg.lbl.gov/>`_ would list.
     """
 
-    name: str
-    pid: int
-    charge: int
-    spin: Spin
-    mass: float
-    strangeness: int = 0
-    charmness: int = 0
-    bottomness: int = 0
-    topness: int = 0
-    baryon_number: int = 0
-    electron_number: int = 0
-    muon_number: int = 0
-    tau_number: int = 0
-    width: Optional[float] = None
-    isospin: Optional[Spin] = None
-    parity: Optional[Parity] = None
-    c_parity: Optional[Parity] = None
-    g_parity: Optional[Parity] = None
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        name: str,
+        pid: int,
+        mass: float,
+        state: QuantumState,
+        width: float = 0.0,
+    ):
+        self.__name = str(name)
+        self.__pid = int(pid)
+        energy = complex(float(mass), float(width))
+        super().__init__(energy, state)
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Particle):
+            return (
+                self.name == other.name
+                and self.pid == other.pid
+                and self.complex_energy == other.complex_energy
+                and self.state == other.state
+            )
+        if isinstance(other, ComplexEnergyState):
+            return super() == other
+        raise NotImplementedError
+
+    def __repr__(self) -> str:
+        return (
+            f"<class {self.__class__.__name__}:"
+            + f" {self.name}, {self.pid},"
+            + f" mass={self.mass},"
+            + f" width={self.width},"
+            + f" state={self.state}"
+        )
+
+    @property
+    def name(self) -> str:
+        return self.__name
+
+    @property
+    def pid(self) -> int:
+        return self.__pid
 
 
 class ParticleCollection(abc.Mapping):
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -146,9 +146,9 @@ class QuantumState(NamedTuple):
     bottomness: int = 0
     topness: int = 0
     baryon_number: int = 0
-    electron_number: int = 0
-    muon_number: int = 0
-    tau_number: int = 0
+    electron_lepton_number: int = 0
+    muon_lepton_number: int = 0
+    tau_lepton_number: int = 0
     parity: Optional[Parity] = None
     c_parity: Optional[Parity] = None
     g_parity: Optional[Parity] = None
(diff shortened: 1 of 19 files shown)


redeboer:
@spflueger Have a look to see if this agrees [with what you had in mind](https://github.com/ComPWA/expertsystem/issues/127#issuecomment-665735782). We also need to think about mutable/immutable (or could leave this up for modification later once these classes are being implemented).

expertsystem/data.py
```diff
@@ -72,7 +72,10 @@ def projection(self) -> float:
 
 
 class Particle(NamedTuple):
-    """Immutable data container for particle info."""
+    """Immutable container of data defining a physical particle.
+
+    Can *only** contain info that the `PDG <http://pdg.lbl.gov/>`_ would list.
```
spflueger, line 77: **only**

expertsystem/data.py
```diff
@@ -126,3 +129,36 @@ def keys(self) -> KeysView[str]:
...
+class QuantumState(NamedTuple):
+    """Collection of properties defining a quantum mechanical state.
+
+    Related to `.Particle`, but can contain more specific quantum state
+    information, such as `.Spin` projection.
+    """
+
+    charge: int
+    spin: Spin
+    mass: float
```
spflueger, line 143: remove the mass here. That will be part of `ComplexEnergyState`

expertsystem/data.py
```diff
@@ -126,3 +129,36 @@ def keys(self) -> KeysView[str]:
...
+    strangeness: int = 0
+    charmness: int = 0
+    bottomness: int = 0
+    topness: int = 0
+    baryon_number: int = 0
+    electron_number: int = 0
+    muon_number: int = 0
+    tau_number: int = 0
```
spflueger, line 151: Maybe you have some good names for these attributes. electron/muon/tau lepton number are the full physics terms

expertsystem/data.py
```diff
@@ -126,3 +129,36 @@ def keys(self) -> KeysView[str]:
...
+    width: Optional[float] = None
+    isospin: Optional[Spin] = None
```
spflueger, line 153: I would move that above strangeness as its a quark flavor qn

expertsystem/data.py
```diff
@@ -126,3 +129,36 @@ def keys(self) -> KeysView[str]:
...
+    mass: float
+    strangeness: int = 0
+    charmness: int = 0
+    bottomness: int = 0
+    topness: int = 0
+    baryon_number: int = 0
+    electron_number: int = 0
+    muon_number: int = 0
+    tau_number: int = 0
+    width: Optional[float] = None
```
spflueger, line 152: Remove. Same reasoning as for mass

expertsystem/data.py
```diff
@@ -126,3 +129,36 @@ def keys(self) -> KeysView[str]:
...
+    parity: Optional[Parity] = None
+    c_parity: Optional[Parity] = None
+    g_parity: Optional[Parity] = None
+
+
+class ComplexEnergyState(NamedTuple):
+    """Pole in the complex energy plane with quantum numbers."""
+
+    position: float
+    width: float
```
spflueger, line 163: To correlate better with the class name, `energy_real` and `energy_imaginary` would be better names in my opinion.

redeboer: What about using Python complex numbers (e.g. [this](https://docs.python.org/3/library/cmath.html))

expertsystem/data.py
redeboer: Start to think we need something more like this:
```python
class LeptonNumbers(NamedTuple):
    electron: int = 0
    muon: int = 0
    tau: int = 0

class QuarkNumbers(NamedTuple):
    strange: int = 0
    charm: int = 0
    bottom: int = 0
    top: int = 0
```

expertsystem/data.py
redeboer: Problem is: the `Particle` class is the starting point, which should contain info from the PDG (where it is called mass and width). For me `ComplexEnergyState` is the mathematical generalisation from that (for internal use and for the output recipe that is used for PWA).

expertsystem/data.py
spflueger: Yeah thats a good idea

expertsystem/data.py
spflueger: Hmm, but what is the gain of using that? Btw isospin is also a quark flavor attribute

redeboer: Just to group things. In practical terms, this would only make it more complicated to construct a `Particle` or `QuantumState` class...

expertsystem/data.py
```diff
@@ -75,36 +78,115 @@ def magnitude(self) -> float:
...
+
+    def __init__(self, energy: complex, state: QuantumState):
+        self.__energy = complex(energy)
+        self.state: QuantumState = state
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, ComplexEnergyState):
+            return self == other
+        if isinstance(other, QuantumState):
+            return self.state == other
```
spflueger, line 122: I think that might just cause troubles. So if only the state should be compared then `.state` should be used in the call

redeboer: bc9976498ec4450626e46f6a4cd7796d6e074246

expertsystem/data.py
```diff
@@ -75,36 +78,113 @@ def magnitude(self) -> float:
...
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Particle):
+            return (
+                self.name == other.name
+                and self.pid == other.pid
+                and self.complex_energy == other.complex_energy
+                and self.state == other.state
+            )
+        if isinstance(other, ComplexEnergyState):
+            return super() == other
```
spflueger, line 167: Same thing here

redeboer: a6fe2fe1566b6375e03fb79f4bf5ca7cdf8254f7

tests/io/test_particle_collection.py
```diff
@@ -70,19 +74,17 @@ def test_yaml_to_xml():
     io.write(yaml_particle_collection, xml_file)
     xml_particle_collection = io.load_particle_collection(xml_file)
     assert xml_particle_collection == yaml_particle_collection
-    dummy_particle = Particle(name="0", pid=0, charge=0, spin=0, mass=0)
+    dummy_particle = Particle(
+        name="0", pid=0, mass=0, state=QuantumState(charge=0, spin=0)
```
spflueger, line 78: we could also make default values for charge (=0) and possibly spin (that might get a bit tricky with the optional projection though)

redeboer: To be honest, I don't understand how this can pass `mypy`. I mean, `spin`, should be `Spin`, not `int`...

redeboer: So the typing error is fixed now (6e756bb8cde5b9e359c39fe852d29c3fe125424e), but this doesn't settle the argument yet. There is indeed something to say for providing default values for all attributes of `QuantumState` (both from a user point of view and a physics point of view).

expertsystem/data.py
```diff
@@ -75,36 +78,114 @@ def magnitude(self) -> float:
...
+    Can **only** contain info that the `PDG <http://pdg.lbl.gov/>`_ would list.
+    """
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        name: str,
+        pid: int,
+        mass: float,
+        state: QuantumState,
+        width: float = 0.0,
```
spflueger, line 154: Can you reorder those to group mass and width together

redeboer: Depends on https://github.com/ComPWA/expertsystem/pull/135#discussion_r463944145. Now, `QuantumState` has to be placed before because it requires an argument.

spflueger: True. For `QuantumState` I would leave spin as a required argument, since its hard to chose a reasonable default (`Particle`: Spin(0, None), `ComplexEnergyState`: Spin(0,0)). I'm starting to think, making a spin projection optional was a mistake. It just too error prone as the user can screw up setting the Spin values incorrectly... That means we would have to go back to the completely separate class for `Particle`. What to you think?

redeboer:
To summarise, to problems remaining:

1. Which values to provide default values for (e.g. spin and charge as well?)
2. Make spin projection optional?

We decided to try out [dataclasses](https://docs.python.org/3/library/dataclasses.html) to avoid the need to work with composition only (`NamedTuple`) and allow for more convenient syntax. Note that this would require us to work with Python 3.7, so we need to be sure of the benefits of dataclasses. Stay tuned.

tests/io/test_particle_collection.py
redeboer: @spflueger Check https://github.com/ComPWA/expertsystem/tree/WIP/dataclass-attempt. It's not really functional, but illustrates the idea. The big problem remains the default values, see https://stackoverflow.com/a/51595616.

redeboer: Okay, all has been addressed in 4140b87f26b3c06d9080fa0332f8ee4a98a67330 (thanks @spflueger)

redeboer:
Had to leave implementation classes such as `HasFloatSpin` public, because otherwise intersphinx runs into trouble. At least the order in the API seems ok
https://pwa--135.org.readthedocs.build/projects/expertsystem/en/135/api/expertsystem.data.html#expertsystem.data.HasComplexEnergy

expertsystem/data.py
```diff
@@ -84,29 +99,121 @@ def __hash__(self) -> int:
         return hash(repr(self))
 
 
-class Particle(NamedTuple):
-    """Immutable data container for particle info."""
+@dataclass(frozen=True)
+class HasQuantumNumbers:  # pylint: disable=too-many-instance-attributes
```
spflueger, line 103: I'm personally don't like the class names with the `Has` prefix. To me the name suggests like a boolean.

expertsystem/data.py
```diff
@@ -84,29 +99,121 @@ def __hash__(self) -> int:
...
+    def width(self) -> float:
+        return self.__energy.imag
+
+
+class ComplexEnergyState(HasComplexEnergy):
+    """Pole in the complex energy plane, with quantum numbers."""
+
+    def __init__(self, energy: complex, state: QuantumState):
+        super().__init__(energy)
+        self.state: QuantumState = state
```
spflueger, line 173: To filter out double solutions later on, we will also need the `__eq__` operation here.

expertsystem/data.py
```diff
@@ -84,29 +99,121 @@ def __hash__(self) -> int:
...
+class QuantumState(HasQuantumNumbers, HasSpin):
+    """Contains all quantum numbers unambiguously defining a quantum state."""
+
+
+@dataclass(frozen=True)
+class ParticleQuantumState(HasQuantumNumbers, HasFloatSpin):
+    """Similar to `.QuantumState` but only carrying spin magnitude."""
+
+
+class HasComplexEnergy:
```
spflueger, line 149: Best add `__eq__` here and just call it from the parent classes.

expertsystem/data.py
redeboer: I know, but I wanted to signal usage somehow and couldn't use `_` here, see https://github.com/ComPWA/expertsystem/pull/135#issuecomment-668142352

expertsystem/data.py
redeboer: Better to implement that once there's a need for it, then it's immediately tested as well.

expertsystem/data.py
redeboer: https://github.com/ComPWA/expertsystem/pull/135#discussion_r464848162

expertsystem/data.py
redeboer: Note that these classes are really only to be used in the local scope, so I don't think naming is that crucial here. Also, PascalCase signals this is a class and cannot be a `bool`.

Issue #44, opened 2020-06-03
Replace particle.DATABASE (nested dict) with ParticleCollection


redeboer:
- [x] Convert `particle.DATABASE` from a `dict` into a `ParticleCollection`
- [x] Fix linting errors (fix static checks)
- [x] Fix tests (fix dynamic checks)
- [x] Implement `ParticleCollection.find` (by PID or name)
- [x] Implement `ParticleCollection.find_subset` (by PID or 'fuzzy' name)
- [x] Implement `merge` and `+=` for `ParticleCollection`
- [x] Remove unused and redundant methods:
  - [x] `add_particle_to_database`
  - [x] `find_particle`
  - [x] `get_particle_copy`
  - [x] `write_particle_database`
- [x] Rewrite notebook about the particle database

> I start to think this problem is strongly related to #22. In the end, this module should just serve as a collection of data containers for particle info -- the `load` and `write` methods are related to serialisation to a custom XML or YAML format that not only contains particle info, but also the intensity recipe etc.

> So what I'm currently giving a try is to design some sort of file adapter or recipe module that can be easily expanded, but that for now would only serve as a factory for the `ParticleDatabase` class. This module should also easily be accesible usable by an external package (i.e. Tensorwaves).

_Originally posted by @redeboer in https://github.com/ComPWA/expertsystem/pull/42_

redeboer:
> https://github.com/ComPWA/expertsystem/pull/42#discussion_r434722133 @spflueger 

What I currently have in mind:
1. Have one global instance of a `ParticleDatabase` that is filled with PDG info obtained from the [scikit-hep/particle](https://github.com/scikit-hep/particle) package.
2. Give the user the option to overwrite that the entries in that instance with custom data from some human readable file format (or hard-coded in Python scripts, see  https://github.com/ComPWA/expertsystem/pull/42#issuecomment-638337174). The latter can be separated out into some sort of recipe or I/O module #22 -- after all, we do want the `expertsystem` will need to write out to some sort human readable file format and I think we should keep that I/O implementation together.
3. Populate the graph nodes with particle info from the global instance.

PR #136, opened 2020-07-30
fix: remove particle database reset

Leongrim:


redeboer:
Thanks, actually this also helped remove a bug for me with regard to the ongoing work for #44 :+1:

PR #145, opened 2020-07-31
ci!: fix check typing in tests

### LINKED ISSUE #144: Move mypy config to setup.cfg
`mypy` works better if the configuration is put under `setup.cfg`, not under `tox.ini`. See [here](https://mypy.readthedocs.io/en/latest/config_file.html#config-file). This would allow us to simplify/remove and the developer can just run `mypy tests`.

https://github.com/ComPWA/expertsystem/blob/7d1df34e53dfb419451365e19c0c5cf650b57430/.pre-commit-config.yaml#L55

and 

https://github.com/ComPWA/expertsystem/blob/88807ac9de3e4c3ab33d208551a7d30b3367ae2a/.vscode/settings.json#L26

redeboer:
Closes #144

### DIFF
--- a/tests/data/test_particle.py
+++ b/tests/data/test_particle.py
@@ -13,8 +13,6 @@ def test_parity():
     parity = Parity(+1)
     assert parity == +1
     assert int(parity) == +1
-    with pytest.raises(AttributeError):
-        parity.value = -1
 
 
 def test_spin():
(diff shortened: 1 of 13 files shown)


redeboer:
@spflueger This PR was sparked by https://github.com/ComPWA/expertsystem/pull/135#discussion_r463706543. We should now check whether `mypy` works correctly under `tests`.

tests/data/test_particle.py
```diff
@@ -13,8 +13,6 @@ def test_parity():
     parity = Parity(+1)
     assert parity == +1
     assert int(parity) == +1
-    with pytest.raises(AttributeError):
-        parity.value = -1
```
spflueger, line 17: Why is this removed?

redeboer: Mypy already checks for read-only access. If we do want to test read-only access, we could add `# type: ignore` here.

PR #147, opened 2020-08-03
feat: add type hints topology module

redeboer:


expertsystem/topology/graph.py
```diff
@@ -10,8 +10,8 @@ def __init__(self):
         self.ending_node_id = None
         self.originating_node_id = None
 
-    def __repr__(self):
-        return str((self.ending_node_id, self.originating_node_id))
+    def __repr__(self) -> str:
+        return f"<class 'Edge': {self.ending_node_id} -> {self.originating_node_id}>"
```
spflueger, line 21: Just now I read up a bit on this topic. `__repr__` should ideally represent the complete state of that object so that all information is available once its logged. We could put it in a string form that it can nicely be verified with `eval(repr(x))`.

So this would mean `return f"Edge({self.ending_node_id},{self.originating_node_id})"`. Then you can test 
```python
c = Edge(0,1)
assert eval(repr(c)) == c
```

redeboer: Ah yeah, you're right, this is how Python does it too, internally:
```python
>>> from typing import NamedTuple
>>> 
>>> 
>>> class Structure(NamedTuple):
...     a: int
...     b: str
... 
>>> 
>>> instance = Structure(7, "text")
>>> print(instance)
Structure(a=7, b='text')
>>> 
```

redeboer: The example
```python
f"Edge({self.ending_node_id},{self.originating_node_id})"
```
Won't work, because the `Edge` constructor takes no arguments. Therefore I mimicked the structure that Python does in that case (although there it would also print `id`). For the other classes, I now test accordingly 50b3a0f12ef4444a11584fa6b98423789ee46157.

PR #148, opened 2020-08-03
ci: fix missing commitlint config

redeboer:


commitlint.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  extends: ["@commitlint/config-conventional"],
+  rules: { "scope-case": [2, "always", "upper-case"] },
```
spflueger, line 3: I don't know what `2` means yet. But I think our scope would be module based or? Then I would stick with the python module conventions `"snake-case"`

redeboer: I think this is for the [(optional) scope](https://www.conventionalcommits.org/en/v1.0.0/#specification), which they for some reason want to be upper case. We don't use scope in the commit messages, so I think we can ignore it, unless we really want to restrict it.

spflueger: Yeah, but if we can ignore it then throw out the whole rules section. I just think upper-case does not make sense here and I would rather use snake-case as it would match pythons module naming convention

redeboer: I would use lower-case, as commit conventions also does

redeboer: Let me check some other options, I want to specify allowed scopes (`feat`, `ci`, etc...)

redeboer:
Okay seems that the PR title linter works as wanted now (see title changes above). We do allow scope now, but lower-case.

commitlint.config.js
spflueger: > I would use lower-case, as commit conventions also does
But then you can remove that line since it wont be different than the default already. Like i said before, we should rather think about what a scope means to us and set that accordingly.

redeboer: I'm not sure what the PR name lint hook has as 'intermediate' default values, so I'd rather keep this line in. Ideally, I don't want scope, but if someone decides to use a scope, I'd prefer that to be in lower case (just as the type). For the rest this PR does what I wanted to achieve with it.

spflueger: So you extend from this conventional config, which means you inherit all of its defined settings. Now you overwrite it with the same setting again. I dont see the point in that. I just falsely suggests that you are overwriting a default of that config, but thats not the case. So I would just remove it. Or am I missing something here

redeboer: Ah yes, that's true

redeboer: Okay let's stick with the defaults bd5f127fa91d3d5f72b908975e6a75fc438f8845

PR #151, opened 2020-08-04
refactor!: replace DATABASE with ParticleColl

### LINKED ISSUE #6: Add functionality to modify expert system particle list through Python
Add functionality to easily add and modify particles in the particle list in the python expert system. Also one should think about the structures to use here (particles and particle list) to improve the expert system code. The XML code has to be separated out into a XML related module!

**Estimated effort: 1d** (this is correlated with issue ComPWA/expertsystem#4)

### LINKED ISSUE #44: Replace particle.DATABASE (nested dict) with ParticleCollection

redeboer:
Closes #6
Closes #44 

- [x] Convert `particle.DATABASE` from a `dict` into a `ParticleCollection`
- [x] Convert `Particle` instances back to `dict` where necessary.
- [x] Implement `ParticleCollection.find` (by PID or name)
- [x] Implement `ParticleCollection.find_subset` (by PID or partial name)
- [x] Implement `merge` and `+=` for `ParticleCollection`
- [x] Remove unused and redundant methods:
  - [x] `add_particle_to_database`
  - [x] `find_particle`
  - [x] `get_particle_copy`
  - [x] `write_particle_database`
- [x] Rewrite notebook about the particle database


redeboer:
@Leongrim @sebastianJaeger The particle database has (hopefully) become easier to use. Please have a look!
https://pwa--151.org.readthedocs.build/projects/expertsystem/en/151/usage/particles.html

As `Particle` instances are immutable, it's a bit bothersome to modify particles, see [here](https://pwa--151.org.readthedocs.build/projects/expertsystem/en/151/usage/particles.html#Adding-custom-particle-definitions-through-Python). What do you think? Should some sort of build-from-a-template-particle function be created for that? Or is working through YAML sufficient?

Leongrim:
What you show in the Notebook looks great! And I think to modify a yaml file is much easier than modifing a xml file. How would it look if the `Particle` instance would be used to create a complete new entry? So with qauntum numbers etc. For instance I have a python script where I have a list of particles and there I create the anti particles and add them to the particle `Database`

expertsystem/data.py
```diff
@@ -238,6 +243,53 @@ def __repr__(self) -> str:
...
+            particle_name = search_term
+            return self.__particles[particle_name]
+        if isinstance(search_term, int):
+            pid = search_term
+            search_results = [
+                particle for particle in self.values() if particle.pid == pid
+            ]
+            if len(search_results) == 0:
+                raise LookupError(f"Could not find particle with PID {pid}")
+            if len(search_results) > 1:
```
spflueger, line 269: I believe a 
```python3
if search_results:
.... 
else:
...
```
statement would be more elegant here

redeboer:
https://github.com/ComPWA/expertsystem/pull/151#issuecomment-668587774
> How would it look if the Particle instance would be used to create a complete new entry?

The problem is that you have to do it [like this](https://pwa--151.org.readthedocs.build/projects/expertsystem/en/151/usage/particles.html#Adding-custom-particle-definitions-through-Python) now, because we don't want `Particle` instances to be mutable (so you can do something like `omega.width = 3.14`). So we could only implement something like
```python
create_particle(template=omega, width=3.14)  # returns a new Particle instance
```
and you can chose to set additional properties like `mass`. Question is: how necessary is such a builder facade, given [how it works currently](https://pwa--151.org.readthedocs.build/projects/expertsystem/en/151/usage/particles.html#Adding-custom-particle-definitions-through-Python).

> For instance I have a python script where I have a list of particles and there I create the anti particles and add them to the particle Database

That sounds quite useful actually. I mean, if it's possible to define an antiparticle unambiguously (!), such a function could be implemented. Something like:
```python
assert pi_minus.state.charge == -1
pi_plus = create_anti_particle(pi_minus)
assert pi_plus.state.charge == +1
```
Notice that this function is called `create...` again, because we cannot modify an output `Particle`, only construct.

redeboer:
As for https://github.com/ComPWA/expertsystem/pull/151#pullrequestreview-460803048 and https://github.com/ComPWA/expertsystem/pull/151#issuecomment-668594796, I think this should be moved to another issue/PR, because this will require some design decisions.

expertsystem/data.py
redeboer: I don't think so, the method finalises because it raises an exception. For sure, `pylint` would complain about an `else` here.

spflueger: Ah nvm I misread that code

Issue #152, opened 2020-08-04
Create builder facades for Particle


redeboer:
Current proposal/idea:

- [x] Create a new `Particle` instance from another `Particle` instance (or `QuantumState` or `ComplexEnergyState` instance):
  ```python
  def create_particle(
      template: Particle,
      name: str,
      pid: int,
      charge: Optional[int] = None,
      spin: Optional[float] = None,
      mass: Optional[float] = None,
      width: Optional[float] = None,
      ...
  ) -> Particle:
      if name == template.name:
          raise ValueError(f"Names should be different: {template.name}")
  ```
  Note that the this removes the need to use `state: QuantumState[float]`, see also https://github.com/ComPWA/expertsystem/pull/153#discussion_r465083919.

- [x] Create a `Particle` instance that is the anti-particle of another `Particle` instance. _This is only possible if one can unambiguously define an anti-particle!_ (For instance, is the PID of an antiparticle always its negative?)
  ```python
  def create_anti_particle(particle: Particle) -> Particle:
  ```
  See also https://github.com/ComPWA/expertsystem/pull/151#issuecomment-668587774

- [ ] Describe these functions in the `examples/particles.ipynb` notebook.

PR #153, opened 2020-08-04
refactor: use generic dataclass as QuantumState

spflueger:


tests/io/test_particle_collection.py
```diff
@@ -23,7 +23,7 @@
     pid=443,
     mass=3.0969,
     width=9.29e-05,
-    state=ParticleQuantumState(
+    state=QuantumState[float](
```
redeboer, line 26: Is there a way to provide a 'default' template argument (and, if so, would this be a desired feature)?

expertsystem/data.py
```diff
@@ -122,32 +126,11 @@ class HasQuantumNumbers:  # pylint: disable=too-many-instance-attributes
...
-    """Required to disallow default arguments for `.ParticleQuantumState`."""
-
-    spin: float
-
-
-@dataclass(frozen=True)
-class QuantumState(HasQuantumNumbers, HasSpin):
-    """Contains all quantum numbers unambiguously defining a quantum state."""
-
+class ComplexEnergy:
```
redeboer, line 129: I'm still not sure about the naming here. For one, `ComplexEnergy` is not to be used outside this module, but okay, that's Sphinx issues. I think that `HasComplexEnergy` signals its intend better. It cannot be a function that returns a `bool`, because it's PascalCase.

redeboer: Note also that the description of this class is misleading too.

expertsystem/data.py
```diff
@@ -164,16 +147,27 @@ def mass(self) -> float:
...
         return self.__energy.imag
 
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, ComplexEnergy):
+            return self.complex_energy == other.complex_energy
+        raise NotImplementedError
 
-class ComplexEnergyState(HasComplexEnergy):
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}{self.complex_energy}"
```
redeboer, line 156: This would return something like `"ComplexEnergycomplex(2.5, 0.3)"`. So should be:
```python
return f"{self.__class__.__name__}{self.complex_energy}"
```
(in other cases in this module, omitting the parentheses was fine, because those were tuples).

expertsystem/data.py
```diff
@@ -164,16 +147,27 @@ def mass(self) -> float:
...
         super().__init__(energy)
-        self.state: QuantumState = state
+        self.state: QuantumState[Spin] = state
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}{self.complex_energy, self.state}"
 
 
-class Particle(HasComplexEnergy):
+class Particle(ComplexEnergy):
```
redeboer, line 170: Naming, as mentioned before: "`Particle` is `ComplexEnergy`" would sound confusing.

expertsystem/data.py
```diff
@@ -183,14 +177,14 @@ def __init__(  # pylint: disable=too-many-arguments
...
-        state: ParticleQuantumState,
+        state: QuantumState[float],
         mass: float,
         width: float = 0.0,
     ):
         super().__init__(complex(mass, width))
         self.__name: str = name
         self.__pid: int = pid
-        self.state: ParticleQuantumState = state
+        self.state: QuantumState[float] = state
```
redeboer, line 187: We may want to do a check on the argument here, like `not isinstance(state, Spin)`. Because the user does have to be familiar with the syntax when creating a new particle.

Might be addressed by #152, otherwise, see remark on 'default' template argument.

tests/data/test_particle.py
```diff
@@ -20,7 +20,7 @@
             pid=443,
             mass=3.0969,
             width=9.29e-05,
-            state=ParticleQuantumState(spin=1, charge=0),
+            state=QuantumState[float](spin=1, charge=0),
```
redeboer, line 23: We may want to add a `__repr__` check on `ComplexEnergy` as well, then that typo would have been fished out. Or even better: somehow automate this test by running over all classes in the `data` module that have a `__repr__` definition (not sure if that's possible).

tests/io/test_particle_collection.py
spflueger: Not that I know. In my opinion a default would also not bring much benefit. You can make a "type alias" within a class to avoid retyping the same class type `_ConcreteQuantumState = QuantumState[float]`

expertsystem/data.py
spflueger: I dont understand the naming issue here. Both are PascalCase right. I will try to work with a class decorator here.
```python3
@ComplexEnergy
class ComplexEnergyState():
...
```

redeboer: Oh yes, that would be way cleaner!

expertsystem/data.py
spflueger: Hmm but mypy is already catching those errors. So how can the instance check help?

expertsystem/data.py
redeboer: Btw, not sure what PEP naming conventions are for class decorators, but seeing as `dataclass` is snake_case, I assume it should be `complex_energy`.

expertsystem/data.py
redeboer: `mypy` does not run on notebooks, which is the most likely place where someone wants to construct a new particle outside the framework.

expertsystem/data.py
spflueger: Good point. Somehow it creates `ComplexEnergy(2.5+j0.3)` though and everything is fine. Do you know how those brackets are automatically inserted?

expertsystem/data.py
spflueger: Ok, I totally missed that. But thats a huge problem. Basically the complete UI has to be protected with isinstance calls as well then

expertsystem/data.py
redeboer: Oh interesting. Well that means the `__repr__` of a `complex` is just `(<real>+<imag>j)`.

expertsystem/data.py
redeboer: Yeah, this touches on a classic Python discussion, I guess, that one assumes a certain responsibility with the user. So we could assume the user not to misuse things here, it's just because of this uncommon template syntax that the change for mistakes is larger.
For sure, #152 would simplify things on a notebook side, so could leave it.

expertsystem/data.py
spflueger: But how is `return f"{self.__class__.__name__}{self.complex_energy, self.state}"` for the `ComplexEnergyState` also correct then? Somehow this `{ ,,,, }` syntax is automatically creating a "tuple".

redeboer: No it's not a tuple, `{ ,,,, }`, just tenders the `__repr__` of a `complex`.

tests/data/test_particle.py
spflueger: see 6511864073ed8b79407c616fe1c6d4cf97552760

expertsystem/data.py
spflueger: I couldn't get this to work with a decorator. I guess it should be doable, but is currently beyond my skillset...

spflueger: But overall the idea would be to automatically generate a form of composition for `ComplexEnergy` on the class to decorate.

expertsystem/data.py
spflueger: That would also be resolved with the decorator

expertsystem/data.py
redeboer: Made an issue for it #157

Issue #138, opened 2020-07-30
Fix coverage mismatch Codacy vs local pytest


redeboer:
There are two big issues with test coverage:
1. `pytest-cov` emits weird warnings on Travis. See all Travis logs for this PR and the original description below.
2. The coverage results on Travis are completely different (around 93%) then if you run `pytest-cov` locally (around 74%).

_Original text:_
> **This PR serves more of a way to try out ideas for the problems we have with Codecov, `pytest-cov` and different test coverage when running locally.**
> 
> If `codecov` is run in the `after_success` stage, the Travis job is not marked "failed" if the `codecov` command fails. This is also why we didn't spot an error like the latest job in the `master` branch:
> ```
>  Error running `['/home/travis/virtualenv/python3.7.6/bin/python', '-m', 'coverage', 'combine', '-a']`: Command '['/home/travis/virtualenv/python3.7.6/bin/python', '-m', 'coverage', 'combine', '-a']' returned non-zero exit status 1.
> ```
> https://travis-ci.com/github/ComPWA/expertsystem/jobs/358409668#L436
> 
> Note in the same job, there seems to be a bigger problem with `pytest-cov` in the `script` stage itself:
> ```
> Coverage.py warning: Data file '/home/travis/build/ComPWA/expertsystem/.coverage.travis-job-14c3cd82-38bf-46c1-b6c8-00de339f5bfa.6151.257560' doesn't seem to be a coverage data file: cannot unpack non-iterable NoneType object
> ```
> I suppose this causes the error for `codecov`. In fact, I wonder how we can have coverage at all, and why this coverage differs so much from the coverage when you run `pytest-cov` locally (see benchmark results in #107).

redeboer:
See also https://stackoverflow.com/questions/58602647/pytest-coverage-return-different-coverage-result-when-testing-cli-applications-i

redeboer:
Has been fixed by manually setting the number of cores to 1 in the tests. Cleaner fix will become available with #169

PR #164, opened 2020-08-06
ci: increase minimal test coverage

redeboer:
This PR allows one to quickly ensure with
```bash
tox -e py  # fast tests only
```
that test coverage does not drop significantly.

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -1,0 +1,26 @@
+name: Linkcheck
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  style:
+    name: Check external links
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[doc]
+          sudo apt-get -y install pandoc
+      - name: Check external links
+        working-directory: doc
+        run: make linkcheck
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,7 @@ whitelist_externals =
 commands =
     pytest tests {posargs} \
         -m "not slow" \
-        --cov-fail-under=92.4 \
+        --cov-fail-under=90 \
         --cov-report=html \
         --cov-report=xml \
         --cov=expertsystem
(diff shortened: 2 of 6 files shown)


tox.ini
```diff
@@ -18,7 +18,7 @@ whitelist_externals =
 commands =
     pytest tests {posargs} \
         -m "not slow" \
-        --cov-fail-under=70 \
+        --cov-fail-under=92.4 \
```
spflueger, line 21: why not 90 like in the `codecov.yml`

redeboer: It's a bit of a dirty hack, to be honest. What I want to achieve is that test coverage **does not drop**. But locally, you cannot easily compare test coverage with that of a previous commit on the master branch. So I picked a value super close to the current test coverage.

Note that he test coverage of `tox -e py` is slightly lower than Codecov, because not all tests are run.

spflueger: but then you would have to update that value all the time. Can't we just let codecov handle that?

redeboer: Well, you don't _have_ to. The way I see it is: codecov already handles that, yes, but you can use `tox` to quickly check if the value has dropped. The only potential problem is that `tox` fails when you run things locally and you implemented new functionality, but this simply means you should write more tests.

For the record, I extract these commits from ongoing work where I wanted to do precisely this.

spflueger: You have to adjust if you want the rate to never drop. Personally I think we should also not cling to the principle that the test rate should never drop. The most important part about tests is that they actually test the critical behavior. I could write a lot of bogus tests that increase the coverage but fail to catch problems. Hence I would not take this coverage value too serious, being 90+ is fine in my current view.

redeboer: Okay, let's stick with 90 then. Note that this PR was originally motivated by #138

.github/workflows/linkcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          python-version: 3.7
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[doc]
+          sudo apt-get -y install pandoc
+      - name: Check external links
+        working-directory: doc
+        run: make linkcheck
```
spflueger, line 26: Seems like it just some backdoor to sidestep the linkcheck. In my opinion it just puts unnecessary load on the github actions.

redeboer: No, I moved the linkcheck job from the CI workflow to this separate workflow.

The thing is, if only one of the jobs in a workflow fails, the whole workflow has to be restarted. So actually this decreases the pressure on Actions, because we regularly restart the linkcheck.

redeboer:
Checks cannot pass while `npm` is down, let's keep an eye on the status
[status.npmjs.org](https://status.npmjs.org/)

PR #181, opened 2020-08-07
feat: add Particle builder façades

### LINKED ISSUE #152: Create builder facades for Particle

sebastianJaeger:
Closes #152


redeboer:
@spflueger 
Do you know if the quantum number 'flips' are correct here?
https://github.com/ComPWA/expertsystem/blob/78f435ec52bc0cc2b7e07cb44d56e4f835c71cc8/expertsystem/state/particle.py#L784-L825

redeboer:
@sebastianJaeger I added a ["Closes" keyword](https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords) to the PR description, because I think this PR can round up this issue.

Only thing that needs to be done though is:
- [ ] Add `create_particle` builder function
- [ ] Implement `create_particle` in `create_antiparticle`
- [ ] Describe usage of the new `create_antiparticle` and `create_particle` functions in the [`particles.ipynb` notebook](https://github.com/ComPWA/expertsystem/blob/master/examples/particles.ipynb)

See #152

tests/state/test_particle.py
```diff
@@ -0,0 +1,25 @@
+from expertsystem.state.particle import DATABASE, create_antiparticle
+from expertsystem.ui import load_default_particle_list
+
+
+def test_create_antiparticle():
```
redeboer, line 5: I think a more solid test would be to use `pytest.mark.parametrize` and test several particles.

sebastianJaeger: Done

examples/particles.ipynb
```diff
@@ -149,17 +149,35 @@
...
-    "    mass=omega.mass,\n",
-    "    width=0.01,\n",
-    "    state=omega.state,\n",
-    ")\n",
-    "particle.DATABASE.add(new_omega)\n",
-    "particle.DATABASE[\"omega(782)\"].width"
+    "new_N1650_minus = create_particle(N1650_minus, name=\"N(1650)-_modified\", width=0.2)\n",
+    "\n",
+    "particle.DATABASE.add(new_N1650_minus)\n",
+    "particle.DATABASE[\"N(1650)-_modified\"].width"
```
redeboer, line 157: I would remove the `.width`, so that Jupyter prints the complete new `Particle` instance. But up to you, it may also make the output cell look ugly

spflueger:
> @spflueger
> Do you know if the quantum number 'flips' are correct here?
> https://github.com/ComPWA/expertsystem/blob/78f435ec52bc0cc2b7e07cb44d56e4f835c71cc8/expertsystem/state/particle.py#L784-L825

Why is that functionality needed in the first place? I can't come up with some use case right now.

First one remark about the pid. In my opinion that concept is not good. One problem that I might see is that particles which are their own antiparticles would have two pids, a positive and the negative one. I think it would be much better if we would just name the antiparticle with a string.

Regarding the correctness of the implementation most of it is correct (only internal quantum numbers change sign). I have only a few remarks:
- The P parity is quite tricky and behaves differently for fermions and bosons (https://en.wikipedia.org/wiki/Parity_(physics) and https://en.wikipedia.org/wiki/Intrinsic_parity). Summary would be P(fermion) = -P(antifermion), P(boson) = P(antiboson)
- I believe the C and G parities do not change for antiparticles (based on the fact that only particles which are their own antiparticle have C defined). 
- Also the Isospin magnitude should not be changed but only the projection (I guess thats implemented correctly)

expertsystem/state/particle.py
```diff
@@ -777,3 +780,118 @@ def merge_qn_props(qns_state, qns_particle):
...
+) -> Particle:
+    isospin = None
+    if template_particle.state.isospin:
+        isospin = -template_particle.state.isospin
+    c_parity = None
+    if template_particle.state.c_parity:
+        c_parity = template_particle.state.c_parity
+    g_parity = None
+    if template_particle.state.g_parity:
+        g_parity = template_particle.state.g_parity
```
spflueger, line 865: Those lines can be removed

expertsystem/state/particle.py
```diff
@@ -777,3 +780,118 @@ def merge_qn_props(qns_state, qns_particle):
...
+    )
+    return new_particle
+
+
+def create_antiparticle(
+    template_particle: Particle, new_name: str = None
+) -> Particle:
+    isospin = None
+    if template_particle.state.isospin:
+        isospin = -template_particle.state.isospin
```
spflueger, line 859: Does the negation only flip the projection?

expertsystem/state/particle.py
```diff
@@ -777,3 +780,118 @@ def merge_qn_props(qns_state, qns_particle):
...
+    if template_particle.state.parity:
+        if (
+            template_particle.state.baryon_number
+            or template_particle.state.electron_lepton_number
+            or template_particle.state.muon_lepton_number
+            or template_particle.state.tau_lepton_number
+        ):
+            parity = -template_particle.state.parity
+        else:
+            parity = template_particle.state.parity
```
spflueger, line 876: As far as I know one only has to differentiate between fermions and bosons. Could be simplified in that regard. So check if `state.spin.magnitude.is_integer()`

sebastianJaeger:
> > @spflueger
> > Do you know if the quantum number 'flips' are correct here?
> > https://github.com/ComPWA/expertsystem/blob/78f435ec52bc0cc2b7e07cb44d56e4f835c71cc8/expertsystem/state/particle.py#L784-L825
> 
> Why is that functionality needed in the first place? I can't come up with some use case right now.

A use case could be that you want to create a number of particles and automatically include their antiparticles, so e.g. in my analysis if i wanted to add a bunch of N+ and the corresponding N- resonances.

> 
> First one remark about the pid. In my opinion that concept is not good. One problem that I might see is that particles which are their own antiparticles would have two pids, a positive and the negative one. I think it would be much better if we would just name the antiparticle with a string.

That might in theory be a problem, but on the other hand, i dont see a case where somebody would want to add an antiparticle to a particle that is its own antiparticle in the first place. And for the other cases it would be useful to be able to identifiy the particles by pid.

> 
> Regarding the correctness of the implementation most of it is correct (only internal quantum numbers change sign). I have only a few remarks:
> 
> * The P parity is quite tricky and behaves differently for fermions and bosons (https://en.wikipedia.org/wiki/Parity_(physics) and https://en.wikipedia.org/wiki/Intrinsic_parity). Summary would be P(fermion) = -P(antifermion), P(boson) = P(antiboson)

The only fermions in the framework are baryons and leptons, so from the current implementation of the particle class, i think i chose the only way of identifying these.

> * I believe the C and G parities do not change for antiparticles (based on the fact that only particles which are their own antiparticle have C defined).

I think so to.

> * Also the Isospin magnitude should not be changed but only the projection (I guess thats implemented correctly)

tests/state/test_particle.py
```diff
@@ -0,0 +1,57 @@
...
+)
+def test_create_antiparticle(particle_name, anti_particle_name):
+    template_particle = DATABASE[particle_name]
+    anti_particle = create_antiparticle(template_particle)
+    comparision_particle = DATABASE[anti_particle_name]
+
+    assert anti_particle.pid == comparision_particle.pid
+    assert anti_particle.mass == comparision_particle.mass
+    assert anti_particle.width == comparision_particle.width
+    assert anti_particle.state == comparision_particle.state
```
spflueger, line 56: comparison

expertsystem/state/particle.py
redeboer: Yes, see #180

tests/state/test_particle.py
redeboer: I don't understand why `cspell` didn't fish this one out

expertsystem/state/particle.py
redeboer: The functionality is tested on paritle-anti-particle pairs in the existing `particle_list.yml`
https://github.com/ComPWA/expertsystem/blob/6977da8a7c98aa48f53f4969878e79b95e163cbc/tests/state/test_particle.py#L44-L47
There it was noticed that the parity does 'flip' for leptons.

tests/state/test_particle.py
redeboer: Aha... https://encyclopedia2.thefreedictionary.com/Comparision

redeboer: e3038b1b666bb71fe7c5d1ce6245551bfc82a805

redeboer:
> Why is that functionality needed in the first place? I can't come up with some use case right now.

As to https://github.com/ComPWA/expertsystem/pull/181#issuecomment-670608362, I would second that. The `expertsystem` is becoming more of a general tool to facilitate decay process hypotheses, so you want to offer this kind of flexibility.

PR #188, opened 2020-08-10
feat: add type hints to ui module

redeboer:


expertsystem/topology/graph.py
```diff
@@ -164,7 +168,7 @@ def attach_edges_to_node_outgoing(
             self.edges[edge_id].originating_node_id = node_id
 
     def get_originating_node_list(
-        self, edge_ids: Sequence[int]
+        self, edge_ids: Union[Set[int], Tuple[int, ...]]
```
spflueger, line 171: But I think `Sequence[int]` is the correct type here, since the code would work with any data type that implements the Sequence interface

expertsystem/ui/__init__.py
```diff
@@ -74,9 +101,9 @@ def __init__(  # pylint: disable=too-many-arguments
                 f"Formalism type {formalism_type} not implemented."
                 f" Use {allowed_formalism_types} instead."
             )
-        self.__formalism_type = formalism_type
-        self.number_of_threads = number_of_threads
-        self.propagation_mode = propagation_mode
+        self.__formalism_type = str(formalism_type)
+        self.number_of_threads = int(number_of_threads)
+        self.propagation_mode = str(propagation_mode)
```
spflueger, line 106: Should also the types of those instance attributes be defined?

expertsystem/ui/_system_control.py
```diff
@@ -219,7 +265,11 @@ def _remove_qns_from_graph(
     return graph_copy
 
 
-def _check_equal_ignoring_qns(ref_graph, solutions, ignored_qn_list):
+def _check_equal_ignoring_qns(
+    ref_graph: StateTransitionGraph,
+    solutions: List[StateTransitionGraph],
+    ignored_qn_list: List[str],
```
spflueger, line 271: Those should be `List[Union[StateQuantumNumberNames, InteractionQuantumNumberNames]]`

spflueger (changes requested):
I'm only partially through the review. I thought I give some quick feedback. I will finish it up later. Its quite messy indeed...

expertsystem/ui/__init__.py
redeboer: `mypy` deduces them from the cast

expertsystem/ui/_system_control.py
redeboer: 36b6f19d88e4fad31caf433ea135f3aa4465501c

expertsystem/topology/graph.py
redeboer: Strange, if I do this, `mypy` complains:
```
expertsystem/ui/_system_control.py:579: error: Argument 1 to "get_originating_node_list" of "StateTransitionGraph" has incompatible type "Set[int]"; expected "Sequence[int]"
```
So `Set` is not a `Sequence`?

spflueger: Yeah true. Sequence implies an order. So instead it should be the unordered from of sequence, I guess `Iterable`

redeboer: :heavy_check_mark: 8e8c2cc2caabaeb2c0b914b7d5e038b5bf332a7c

spflueger: Hmm, ok we have to think about this a bit more. I feel like this has to be a `Set` (since the ids have to be unique)

spflueger: So what I'm trying to say is that we have to like modify the types of the nodes and edge of a graph in the future. Since this PR is just about adding types on the current code, this issue has to be tackled separately. (see #226)

redeboer: Yes, that's better

tests/unit/amplitude/test_parity_prefactor.py
```diff
@@ -81,6 +81,8 @@ def test_parity_prefactor(
         ]
         assert len(in_edge) == 1
         node_id = solution.edges[in_edge[0]].ending_node_id
+        if node_id is None:
+            continue
```
spflueger, line 85: What is this good for?

expertsystem/ui/_system_control.py
```diff
@@ -170,9 +217,9 @@ def remove_duplicate_solutions(
     return filtered_results
 
 
-def _remove_qns_from_graph(
-    graph, qn_list
-):  # pylint: disable=too-many-branches
+def _remove_qns_from_graph(  # pylint: disable=too-many-branches
+    graph: StateTransitionGraph, qn_list: list
```
spflueger, line 221: If you want it to be precise qn_list should also be a `Union[InteractionQuantumNumberNames, StateQuantumNumberNames, ParticlePropertyNames]`

tests/unit/amplitude/test_parity_prefactor.py
redeboer: To make `mypy` happy. Because `node_id` has `Optional[int]` type prior to line 84.

expertsystem/ui/_system_control.py
redeboer: Could be that I tried, but that things blew up elsewhere. So I practically stated with `List[Any]` here.

tests/unit/amplitude/test_parity_prefactor.py
spflueger: Ok, would asserting also work? That would make the test a bit stronger. So assert node_id is of type int.

expertsystem/ui/_system_control.py
spflueger: Ok, if it doesnt work just leave it like this. Having the types pinned down more than Any is already useful. Through the refactoring everything will change anyways, so we do not need to be super critical here.

redeboer: Ok works actually 2019d1810c4cfc12db6e9ffa7d1d27fb3cd8a115

tests/unit/amplitude/test_parity_prefactor.py
redeboer: Not sure, but why not instead of a `continue`?

spflueger: Hmm it should never happen. So if you continue, you would just step over this error if it actually does happen because of some bug.

spflueger: So `assert isinstance(node_id, int)` does not work?

spflueger (approved):
Good work

Issue #189, opened 2020-08-10
Switch to PDG as source of truth


redeboer:
- [ ] Use the new implementation introduced by #23 to load the `particle.DATABASE`.
- [ ] Put more emphasis on `allowed_intermediate_particles`: if not setting this argument, there will be a huge number of solutions
- [ ] **Rewrite all notebooks and tests** to use a `allowed_intermediate_particles`

Issue #33, opened 2020-05-15
Remove particle_list.*ml files


spflueger:
Keep source of truth to one file. 
Maybe even better: get default particles from the hep particle package.

redeboer:
https://github.com/ComPWA/expertsystem/issues/23

PR #190, opened 2020-08-10
feat: create ParticleCollection from PDG

### LINKED ISSUE #23: Get particle info from PDG

### LINKED ISSUE #199: Decide whether to choose a convention for parity of fermions
This issue came up while working on #190, where PDG info is imported from the scikit-hep/particle package into Particle instances. That package leaves parity undefined for leptons , which led us to reconsider our current 'source of truth', particle_list.yml (see also #189). That list indeed defines parity as +1 for fermions, -1 for anti-fermions.

Purely in technical terms, we now need to either (1) let the PDG adapter attribute +1 parity to fermion Particle instances or (2) leave parity undefined in Particle fermions and attribute the parity once particle properties are propagated through the StateTransitionGraphs.

See also https://github.com/ComPWA/expertsystem/pull/181#issuecomment-670602932

Leongrim:
Added the functionality to create a particle list from [scikit-hep `particle` package](https://github.com/scikit-hep/particle) 

Closes #23
Closes #199 (63b0c11cff2f0833c7489dc20fd2ff9bf56e2081)

## To-do list
- [x] Improve quark content --> isospin projection
- [x] Fix isospin bug for K(S/L)0
- [x] Fix antibaryon parity bug in `particle` package --> https://github.com/scikit-hep/particle/issues/250
- [x] Find solution for `"Maybe non-qq"` --> https://github.com/ComPWA/expertsystem/pull/190#discussion_r471418353
- [x] Match floating point precision
- [x] Compare all entries with `particle_list.yml` --> #202 #203


expertsystem/io/__init__.py
```diff
@@ -9,11 +9,14 @@
...
 from expertsystem.data import ParticleCollection
 
+from . import _pdg
 from . import xml
 from . import yaml
 
 
-def load_particle_collection(filename: str) -> ParticleCollection:
+def load_particle_collection(filename: str = "") -> ParticleCollection:
+    if not filename:
```
redeboer, line 18: I'm not sure if we should have this kind of hidden behaviour. Instead, probably better to keep the `load_pdg` function separate, and make it public. (Yes, I know it was me who implemented it like this...)

What do you think @spflueger?

expertsystem/io/_pdg.py
```diff
@@ -0,0 +1,129 @@
...
+)
+
+
+# cspell:ignore pdgid
+def __get_lepton_qn(pdg_particle: PdgDatabase) -> tuple:
+    electron_lepton_number = 0
+    muon_lepton_number = 0
+    tau_lepton_number = 0
+    if pdg_particle.pdgid.is_lepton:
+        if "e" in pdg_particle.name:
```
redeboer, line 23: This will cause problems: I guess there are several particles that have an `"e"` in their name but are not a lepton. Probably better to use something like `endswith` (or starts, not sure).

expertsystem/io/_pdg.py
```diff
@@ -0,0 +1,129 @@
...
+            electron_lepton_number = pdg_particle.pdgid / abs(
```
redeboer, line 24: Nice trick :) But better to compute it once, right after the `if pdg_particle.pdgid.is_lepton`

expertsystem/io/_pdg.py
```diff
@@ -0,0 +1,129 @@
...
+                pdg_particle.pdgid
+            )
+        if "mu" in pdg_particle.name:
```
redeboer, line 27: `elif`

expertsystem/io/_pdg.py
```diff
@@ -0,0 +1,129 @@
...
+    charmness = 0
+    bottomness = 0
+    topness = 0
+    baryonnumber = (
+        pdg_particle.pdgid
+        / abs(pdg_particle.pdgid)
+        * pdg_particle.pdgid.is_baryon
+    )
+    if pdg_particle.pdgid.is_hadron:
+        for quark in pdg_particle.quarks:
```
redeboer, line 50: Better to create a separate function, something like:
```python
def __compute_quark_numbers(quark_string: str) -> Tuple[int, int, int, int]:
    ...
```
It's still not clean, but the smaller the return tuple, the smaller the risk for bugs.

expertsystem/io/_pdg.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+        if "mu" in pdg_particle.name:
+            muon_lepton_number = pdg_particle.pdgid / abs(pdg_particle.pdgid)
+        if "tau" in pdg_particle.name:
+            tau_lepton_number = pdg_particle.pdgid / abs(pdg_particle.pdgid)
+
+    return electron_lepton_number, muon_lepton_number, tau_lepton_number
+
+
+def __get_hadron_qn(
```
redeboer, line 35: I try to reserve the term `get` for functions that don't do more than 'getting', and use something like `compute` or `create` to signal when a function does a computation.

expertsystem/io/_pdg.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def __create_parity(parity_enum: enums.Parity) -> Optional[Parity]:
+    if parity_enum in [enums.Parity.o, enums.Parity.u]:
+        return None
+    return Parity(parity_enum)
+
+
+def __create_spin(magnitude: float, projection: float) -> Optional[Spin]:
+    if magnitude is not None:
```
redeboer, line 88: Would reverse the order (to avoid the `not`).

expertsystem/io/_pdg.py
```diff
@@ -0,0 +1,129 @@
...
+        return Spin(magnitude, projection)
+    return None
+
+
+def load_pdg() -> ParticleCollection:
+    all_pdg_particles = PdgDatabase.findall(
+        lambda item: item.charge.is_integer()  # remove quarks
+        and item.J is not None  # remove new physics and nucleus
```
redeboer, line 96: nuclei :grimacing:

tests/io/test_particle_collection.py
```diff
@@ -126,3 +126,13 @@ def test_find_subset():
         assert len(search_result) == 1
         assert gamma_from_subset.pid == 22
         assert gamma_from_subset is particle.DATABASE["gamma"]
+
+
+def test_pdg():
+    pdg_particle_collection = io.load_particle_collection()
+    # io.write(pdg_particle_collection, "pdg.yml")
+    assert len(pdg_particle_collection) == 541
+    d_meson = pdg_particle_collection["D0"]
```
redeboer, line 135: For the eventual #189, it would be better if we compare to the existing definitions in `particle_list.yml`. This would also be a more thorough test for `load_pdg`, for instance, it would have fished out the problem with (see comment elsewhere):
https://github.com/ComPWA/expertsystem/blob/cac9256a9985114c27ab7752e1c235798a5c3e7a/expertsystem/io/_pdg.py#L23

Note that this requires matching the particle names in `particle_list.yml` with those used in the `particle` package. That can be done through:
1. a `pytest.mark.parametrize` decorator, just like the 'mapping' used by @sebastianJaeger here:
 
 https://github.com/ComPWA/expertsystem/blob/29d82ec786c09028a3ef792146de43b731a15b4a/tests/state/test_particle.py#L44-L48

2. Renaming the entries of `particle_list.yml` accordingly. But this also requires modifying several tests and possibly the notebooks.

For sure it is worth the effort, because need to ensure that the quantum numbers of `particle_list.yml` are the same and that the `load_pdg` function works correctly.

expertsystem/io/_pdg.py
Leongrim: the `if pdg_particle.pdgid.is_lepton:` will get rid of particles that are no leptons 😉

expertsystem/io/_pdg.py
Leongrim: I´m not sure what you mean, but the only thing in this function that is not calculated from the quarks is the baryon number.

expertsystem/io/_pdg.py
redeboer: Ah yeah good point

expertsystem/io/_pdg.py
redeboer: This function would only calculate `strangeness`, `charmness`, `bottomness`, and `topness` (return as a 4-tuple). So you extract this function from the `__get_hadron_qn` and let `__get_hadron_qn` only return `baryonnumber` and `spin_projection` (probably need to rename it).

The aim is just to keep functions small and reduce the size of the returned tuples.

expertsystem/io/__init__.py
spflueger: Agreed. Then its more clear from the code whats happening

tests/io/test_particle_collection.py
redeboer: Working on a complete comparison, seems several particle definitions are not the same as what is imported from the PDG (c1ae4e32997bb62fed367c534195e21553ad7575)

redeboer:
There's still a bug in in the computation of isospin projection from quark content. This becomes clear from `"K_S0"`, which now gets `isospin=Spin(0.5, 0.0)`.

expertsystem/io/_pdg.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def __compute_isospin_projection(pdg_particle: PdgDatabase) -> float:
+    spin_projection = 0.0
+    if pdg_particle.pdgid.is_hadron:
+        for quark in pdg_particle.quarks:
+            if quark in ["u", "D"]:
+                spin_projection += 0.5
+            elif quark in ["U", "d"]:
+                spin_projection -= 0.5
```
spflueger, line 136: Ok there is a problem here, originating from the `particle` package side. For example the particles `a0(980)0/+/-` all have a quark content defined as `'Maybe non-qQ'`. So in other words its not set... In my honest opinion, this make the `particle` package pretty untrustworthy. As a workaround, you could use the Gellmann Nishijima formula to calculate the correct isospin projection. https://en.wikipedia.org/wiki/Gell-Mann%E2%80%93Nishijima_formula. @redeboer is already writing a test to check the consistency of particle definitions. This will also use this formula, so you could try to reuse that.

redeboer: @spflueger There are valid reasons to leave the quark content vague or `"Maybe non-qQ"`, see
https://pdg.lbl.gov/2020/reviews/rpp2020-rev-non-qqbar-mesons.pdf

For now, I use a mapping to 'fix' the isospin projections
https://github.com/ComPWA/expertsystem/blob/da271944af63ddd209198a463937b4c814a7fa8f/expertsystem/io/_pdg.py#L138-L146
It's just to sketch the idea, because there are 12 particles with `"Maybe non-qQ"`, see
```python
particle.Particle.findall(lambda p: "qq" in p.quarks.lower())
```

Other unconventional quark content strings are now parsed with regex
https://github.com/ComPWA/expertsystem/blob/da271944af63ddd209198a463937b4c814a7fa8f/expertsystem/io/_pdg.py#L165-L169

I first want to reorganise the `test_particle_collection` tests a bit in another PR, meanwhile we can think whether there is a more stable solution to the `"Maybe non-qQ"` problem.

redeboer: For the record, it concerns these:
- [`f(0)(500)`](https://pdg.lbl.gov/2020/listings/rpp2020-list-f0-500.pdf)
- [`f(0)(980)`](https://pdg.lbl.gov/2020/listings/rpp2020-list-f0-980.pdf)
- [`a(0)(980)0+-`](https://pdg.lbl.gov/2020/listings/rpp2020-list-a0-980.pdf)
- [`pi(1)(1400)0+-`](https://pdg.lbl.gov/2020/listings/rpp2020-list-pi1-1400.pdf)
- [`f(0)(1500)`](https://pdg.lbl.gov/2020/listings/rpp2020-list-f0-1500.pdf)
- [`pi(1)(1600)0+-`](https://pdg.lbl.gov/2020/listings/rpp2020-list-pi1-1600.pdf)

Note that the 𝜋₁ states are exotic (𝐽^𝐶𝑃 = 1⁻⁺), so I guess we have to exclude these from the import.

spflueger: @redeboer Thanks for the additional infos. I'm aware of the reason behind "maybe non-qqbar". However all of the conserved quantum numbers in reactions can precisely be specified, regardless of the precise quark content. So I would make more sense if the `particle` package would provide this info.

Regarding the 𝜋₁ state. "Exotic" states are particularly interesting to handle in reaction, so they should be definitely included as well. Just to clarify, those states are just not "constructable" in the simple meson (quark antiquark) picture. But they are perfectly fine to use inside the expertsystem itself

redeboer: Should be solved with the latest commits. Last thing I want to tackle cleanly is the mismatch in floating point precision (some values from the `particle` package come in as `...0000002`, for instance).

redeboer:
Alrighty, I think the major issues are solved now. Not completely satisfied with the current implementation, but I think we're getting close now.

@spflueger Ideas?

tests/unit/data/test_particle_collection.py
```diff
@@ -11,7 +11,7 @@ def test_find(particle_database):
     assert f2_1950.mass == 1.936
     phi = particle_database.find("phi(1020)")
     assert phi.pid == 333
-    assert phi.width == 0.004249
+    assert phi.width == 0.004248999999999999
```
spflueger, line 14: Have a look at https://docs.pytest.org/en/latest/reference.html#pytest-approx

redeboer: I wonder how that would work for a class with several attributes. I did it with these strings, because I assume (#33) these particle lists will disappear eventually anyway. At first I also rounded the floats upon imports, and even tried out `Decimal`, but all of those introduce hidden behaviour.

redeboer: See 170a905da2d38760365d4d2188b039ddaee87a19

expertsystem/io/_pdg.py
```diff
@@ -0,0 +1,171 @@
+"""Create a `.ParticleCollection` instance from PDG info."""
```
spflueger, line 1: Why is this module private?

expertsystem/io/__init__.py
```diff
@@ -24,6 +20,15 @@ def load_particle_collection(filename: str) -> ParticleCollection:
     )
 
 
+def load_pdg() -> ParticleCollection:
+    """Create a `.ParticleCollection` with all entries from the PDG.
+
+    PDG info is imported from the `scikit-hep/particle
+    <https://github.com/scikit-hep/particle/blob/master/README.rst>`_ package.
+    """
+    return _pdg.load_pdg()
```
spflueger, line 29: Why not just make the pdg module public and use `load_pdg()` directly from there?

tests/unit/data/test_particle_collection.py
```diff
@@ -40,14 +42,15 @@ def test_exceptions(particle_database):
     new_particle = create_particle(
         template_particle=particle_database["gamma"], name="gamma_new"
     )
-    particle_database += new_particle
+    dummy_database = deepcopy(particle_database)
+    dummy_database += new_particle
```
spflueger, line 46: Hmm I thought using the `scope="class"` for this fixture was the preferred solution?

expertsystem/io/_pdg.py
redeboer: To hide implementation details. I want to keep the `io.__init__.py` clean because it's accumulating a lot of functionality (probably even #216).

expertsystem/io/__init__.py
redeboer: https://github.com/ComPWA/expertsystem/pull/190#discussion_r472318491

expertsystem/io/_pdg.py
spflueger: But everything is already private in there except `load_pdg()`. So basically you want to be able to do `io.load_pdg()` instead of `io.pdg.load_pdg()`? And what is the point of hiding the implementation detail of a public function?

redeboer: It's about how the Sphinx API renders

redeboer: https://pwa--190.org.readthedocs.build/projects/expertsystem/en/190/api/expertsystem.io.html#module-expertsystem.io

spflueger: I see. But why not move the `load_pdg()` function from `_pdg.py` to `__init__.py`?

redeboer: Have a look e962dcd785d68b566ebe1f1e6245f347dda0dda7

tests/unit/data/test_particle_collection.py
redeboer: Ok, testing whether it works as expected

redeboer: Nope, also using
```python
@pytest.fixture(scope="class")
def particle_database() -> ParticleCollection:
    ...
```
instead of
https://github.com/ComPWA/expertsystem/blob/21d0cf79a4bec0f88ea5b69f3a5ead79c9187e11/tests/unit/conftest.py#L20-L23

also kills the channel tests. So seems that the `deepcopy` is the safest solution, even if dirty. Something to keep in mind for #219.

spflueger: Ah, I get it now. Even thought the fixture will be executed and destroyed many times (on the class level scope), the  DATABASE is a global still persists the old state.
So in order to really have fixture we need to get rid of this DATABASE global. In this case it would be enough to just load a `ParticleCollection` from the pdg or yaml file and just return that.

redeboer: 71fd1fb73da35186b2bc0034b6fad24342b3ab1b would have been easier... :man_facepalming:

redeboer (approved):

expertsystem/io/_pdg.py
spflueger: Actually I prefer the previous way :grimacing:

spflueger (approved):

expertsystem/io/_pdg.py
redeboer: No problem :sweat_smile: 325f62bf801c87bf81615bd534fa94cdd5de278e

PR #191, opened 2020-08-10
ci: remove mypy module ignores

redeboer:
Previously, we made the mistage to use the `mypy` option `ignore_errors` to completely ignore certain modules. This also removes checks on functions that use types from those ignored module. Using `disallow_untyped_defs = False` would have been sufficient.

Note that this change pointed out some mistakes in the `create_particle` builder functions. These have been fixed.

### DIFF
--- a/expertsystem/state/particle.py
+++ b/expertsystem/state/particle.py
@@ -23,6 +23,7 @@ from numpy import arange
 
 from expertsystem import io
 from expertsystem.data import (
+    Parity,
     Particle,
     ParticleCollection,
     QuantumState,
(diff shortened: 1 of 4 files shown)


expertsystem/state/particle.py
```diff
@@ -788,23 +789,23 @@ def create_particle(  # pylint: disable=too-many-arguments,too-many-locals
...
+    bottomness: Optional[int] = None,
+    topness: Optional[int] = None,
+    baryon_number: Optional[int] = None,
+    electron_lepton_number: Optional[int] = None,
+    muon_lepton_number: Optional[int] = None,
+    tau_lepton_number: Optional[int] = None,
+    parity: Optional[int] = None,
+    c_parity: Optional[int] = None,
+    g_parity: Optional[int] = None,
+    state: Optional[QuantumState[float]] = None,
```
redeboer, line 806: @sebastianJaeger Type hints are now checked for the `expertsystem.state.particle` module as well, so I had to fix the type hints here and there.

spflueger: Wah, how did I miss those types in the last PR...

PR #193, opened 2020-08-10
fix!: distinguish isospin 0 and None

redeboer:
Set `isospin=Spin(...)` for all hadrons and `None` (undefined) for leptons and bosons.

### DIFF
--- a/expertsystem/io/xml/_dump.py
+++ b/expertsystem/io/xml/_dump.py
@@ -87,7 +87,7 @@ def _to_quantum_number_list(
     for type_name, instance in conversion_map.items():
         if instance is None:
             continue
-        if type_name not in ["Charge", "Spin"] and instance == 0:
+        if type_name not in ["Charge", "Spin", "IsoSpin"] and instance == 0:
             continue
         definition = _qn_to_dict(instance, type_name)
         output.append(definition)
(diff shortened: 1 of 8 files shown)


expertsystem/io/xml/_dump.py
```diff
@@ -87,7 +87,7 @@ def _to_quantum_number_list(
     for type_name, instance in conversion_map.items():
         if instance is None:
             continue
-        if type_name not in ["Charge", "Spin"] and instance == 0:
+        if type_name not in ["Charge", "Spin", "IsoSpin"] and instance == 0:
```
spflueger, line 90: Why only Charge and Spins are checked here?

redeboer: In the former situation:
`Charge` and `Spin` were always written out, even if they were were 0 (because it's a required property, also according to the JSON schema)

Now:
We want to write out `IsoSpin` as well _if it is defined_, but not if it is `None`.

It's a bit more technical, because the `Spin` has an `__eq__` method that can compare with `int` and `float`, which messes up the logic in the loop here. But I would say these `_dump` modules a mess anyway, I don't want to spend too much time cleaning that up.

spflueger: Yeah for me it was more about why not the other quantum numbers as well. But I guess you only care about isospin like the PR title suggests.

redeboer: Ah yeah. What this loop does (among others) is not serialise things like baryon number if they are 0, as to keep the YAML/XML small.

PR #200, opened 2020-08-11
feat: check spin magnitude+projection consistency

### LINKED ISSUE #205: Check consistency of particle definitions
Use Gellmann Nishijima Rule and perhaps more to verify particles are defined consistently and immediately pick up wrong definitions

redeboer:
Inpsired by https://github.com/ComPWA/expertsystem/pull/190#issuecomment-672129069

This check makes sure that the following examples are not allowed anymore:
```python
Spin(0.3, 0.3)
Spin(1.0, 0.5)
Spin(0.5, 0.0)
```

Since the previous Gellmann-Nishijima check on `Particle` definitions only verifies the consistency between certain quantum, the "type" of certain quantum numbers has to also be verified. This is that the charge of a particle is integral, (.. -1, 0, 1,...) and that the Isospin of a particle has an appropriate spin and projection setting. The latter is handled with this PR

Closes #205


redeboer:
@spflueger 
The tests will fail, because there's a problem: our definitions of K0S and K0S have spin projection 0.0, but magnitude 0.5. But I don't think they actually carry spin projection (K0 and K0bar do).

expertsystem/data.py
```diff
@@ -63,11 +63,20 @@ class Spin(abc.Hashable):
...
+        if magnitude % 0.5 != 0.0:
+            raise ValueError(
+                f"Spin magnitude {magnitude} has to be a multitude of 0.5"
+            )
         if abs(projection) > magnitude:
             raise ValueError(
                 "Spin projection cannot be larger than its magnitude:\n"
                 f"  {projection} > {magnitude}"
             )
+        if (projection - magnitude) % 1.0 != 0.0:
```
spflueger, line 75: I like the simplicity of this check. You could also consider using `is_integer`

expertsystem/data.py
```diff
@@ -63,11 +63,20 @@ class Spin(abc.Hashable):
     def __init__(self, magnitude: float, projection: float) -> None:
         magnitude = float(magnitude)
         projection = float(projection)
+        if magnitude % 0.5 != 0.0:
```
spflueger, line 66: Also add a check for `magnitude >= 0.0`

expertsystem/data.py
redeboer: Ah yeah good one

expertsystem/data.py
redeboer: Not needed because of exception upon `abs(projection) > magnitude`

redeboer:
@spflueger 
I think the check
https://github.com/ComPWA/expertsystem/blob/b0dc5dfdd15fc87918a4d87834dbefbff599130d/expertsystem/data.py#L75
will have to be removed until #7 is resolved.

Question remains: do we just set isospin projection to `0.0` if undefined? This will have to be decided about also with regard to #190.

expertsystem/data.py
spflueger: True, tricky :smile:

spflueger:
So we only have two options. 
1. Either we add the possibility to have a spin projection of `None` for an Isospin as well
2. Or we define linear combination states. This would mean `Particle` are strong eigenstates (well defined quark flavor). And a `ComplexEnergyState` has to get the ability to be a linear combination of states. Either adding a new class etc or add that functionality into `ComplexEnergState`

Option 1 would be a fast implemented solution. Option 2 is better imo as it follows physics accurately.

redeboer:
> 2. Or we define linear combination states. This would mean Particle are strong eigenstates (well defined quark flavor). And a ComplexEnergyState has to get the ability to be a linear combination of states. Either adding a new class etc or add that functionality into ComplexEnergState

I also prefer this solution, but yes, this would would require new functionality.

That leaves 1. for which I'll make a separate issue

spflueger:
> > 1. Or we define linear combination states. This would mean Particle are strong eigenstates (well defined quark flavor). And a ComplexEnergyState has to get the ability to be a linear combination of states. Either adding a new class etc or add that functionality into ComplexEnergState
> 
> I also prefer this solution, but yes, this would would require new functionality.
> 
> That leaves 1. for which I'll make a separate issue

Check out my update of https://github.com/ComPWA/expertsystem/pull/200#issuecomment-672735338

redeboer:
@spflueger The tests fail because of how the default domains generate non-physical spin instances.

spflueger:
Ok do you need my help on this or you know how to solve it?

redeboer:
I'll check it out tomorrow. But I have a feeling this opened up another of Pandora's boxes :sweat_smile:

spflueger:
Ok, small update regarding this issue. So it your fear of opening a Pandora's box was indeed justified :laughing: 

The root of the problem is the mixing of canonical and helicity formalism. More precisely the `InteractionQuantumNumber.S` is of type Spin, so has a magnitude and projection. This is great and how it supposed to be. However in the helicity formalism `S` (as well as `L`) are not defined. Not having `S` and `L` available, in turn makes a couple of rules unusable as well (e.g. `CParityConservation`). You can sidestep this issue by defining `S` and `L` anyways, while "ignoring" the projection as only the magnitude is of importance for these rules. Currently the spin projections are just set to zero to avoid duplicate indistinguishable solutions.

How to solve this nicely is quite tough. Right now I only have two ideas:
1. instead of setting the projection to 0 set it equal to the magnitude (and go one from there)
2. allow the type of `S` and `L` to be either `Spin` or `float`

Fastest for now would be option 1. I would have to give this more though, on what is better long term...

redeboer:
So, for now, we decided to stick with
> instead of setting the projection to 0 set it equal to the magnitude (and go one from there)

I gave it a try in 81dfa289e5700dc890781232b6a019a1229958d0, but seems that there are some other things still going wrong under the surface.

redeboer:
Patch coverage fixed now, :ballot_box_with_check: from my side

@spflueger Does this close #205?

spflueger:
> Patch coverage fixed now,  from my side
> 
> @spflueger Does this close #205?

Yeah, I update the description of this issue

Issue #197, opened 2020-08-11
Isolate all particle interfaces to a separate submodule


redeboer:
Classes and functions like `ParticleCollection` and `load_pdg` (#190) start to become more standalone and can be isolated to a separate module.

Suggested name: `expertsystem.particles`.

Notice that this removes the need for a `expertsystem.data` submodule, although this may again be affected by #22.

redeboer:
Closed by #314

PR #202, opened 2020-08-12
fix!: update PDG values in particle_list

redeboer:


redeboer:
@spflueger I'm only worried about 
6226f16

spflueger:
> @spflueger I'm only worried about [6226f16](https://github.com/ComPWA/expertsystem/commit/6226f16f79eefb158e5df67d58a443de24cc6136) because of the required changes in [`tests/channels/test_d0_to_kskpkm.py`](https://github.com/ComPWA/expertsystem/commit/6226f16f79eefb158e5df67d58a443de24cc6136#diff-246144dfae6d956da3de29ad8ccbbe81)

Hmm I can take a look.

expertsystem/particle_list.yml
```diff
@@ -280,28 +291,30 @@ ParticleList:
...
-    PID: 9000211
-    Mass: 0.994
+    PID: -9000211
+    Mass: 0.98
+    Width: 0.075
     QuantumNumbers:
       Spin: 0
       Charge: -1
       Parity: +1
-      IsoSpin: { Value: 1, Projection: -1 }
```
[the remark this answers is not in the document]
redeboer: Hm, then we have to revisit the quark content -> spin projection rules defined by #190. That's where I got these numbers from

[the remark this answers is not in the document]
spflueger: Could you clarify? I cannot follow

spflueger:
So implementing the test as requested by #205 is to tedious currently. Once we switch the rules to use QuantumState like objects it will be much easier. So I would postpone #205 until later

redeboer:
> So implementing the test as requested by #205 is to tedious currently.

Why is it tedious? I propose to:
1. add a `hypercharge` property to `QuantumState` (works for both `QuantumState[float]` and `QuantumState[Spin]`
2. do the Gell-Mann–Nishijima formula check in the constructor of `Particle`.

wgradl:
> A faulty definition of the charged a0(980) particles make the D decay test fail. Currently a warning message is print out in such a case (quantum number solutions are found, but no matching particle can be inserted, due to a wrongly defined particle). I opened issue #205 to avoid having faulty particles in general. It would be a good options to include such a test in this PR, to ensure all particle definitions are correct.
> 
> >      QuantumNumbers:
>        Spin: 0
>        Charge: -1
>        Parity: +1
> -      IsoSpin: { Value: 1, Projection: -1 }
> 
> The projection -1 was correct
> 
> >      QuantumNumbers:
>        Spin: 0
>        Charge: 1
>        Parity: +1
> -      IsoSpin: { Value: 1, Projection: 1 }
> 
> The projection +1 was correct. This is the reason why the `D0 -> K0 K+ K-` only finds 3 solutions now.

hmmm ... consider the decay D0 --> a0(980)+ K-  on the quark level:
  [c ubar] --> [s ubar] W+ --> [s ubar] [u dbar] ~ K- a0+

the a0+ then decays like this:
  [u dbar] --> [u (sbar s) dbar] --> [u sbar] [s dbar] ~ K+ K0_bar

so, the correct decay to consider would be D0 --> K0_bar K+ K-.  No?

-- Wolfgang




-- 
Prof Dr Wolfgang Gradl        University of Mainz 
BES-III, BABAR & A2@MAMI      Institut für Kernphysik
Tel. +49-6131-39-25871

spflueger:
> > So implementing the test as requested by #205 is to tedious currently.
> 
> Why is it tedious? I propose to:
> 

You have to convert the Particle to a dict of Enums etc
> 1. add a `hypercharge` property to `QuantumState` (works for both `QuantumState[float]` and `QuantumState[Spin]`
Since hypercharge can be calculated from the other QN already, this info is redundant and just adds one more point of inconsistency. Hence I oppose that
> 2. do the Gell-Mann–Nishijima formula check in the constructor of `Particle`.
That is a good point.

spflueger:
> hmmm ... consider the decay D0 --> a0(980)+ K- on the quark level: [c ubar] --> [s ubar] W+ --> [s ubar] [u dbar] ~ K- a0+ the a0+ then decays like this: [u dbar] --> [u (sbar s) dbar] --> [u sbar] [s dbar] ~ K+ K0_bar so, the correct decay to consider would be D0 --> K0_bar K+ K-. No?

@wgradl Maybe there is slight misunderstanding. We are currently testing `D0 --> K0_bar K+ K-`. Just the isospin of a0(980)+/- were defined incorrectly in this PR, which made the test fail. So everything is ok.

wgradl:
> > hmmm ... consider the decay D0 --> a0(980)+ K- on the quark level: [c ubar] --> [s ubar] W+ --> [s ubar] [u dbar] ~ K- a0+ the a0+ then decays like this: [u dbar] --> [u (sbar s) dbar] --> [u sbar] [s dbar] ~ K+ K0_bar so, the correct decay to consider would be D0 --> K0_bar K+ K-. No?
> 
> @wgradl Maybe there is slight misunderstanding. We are currently testing `D0 --> K0_bar K+ K-`. Just the isospin of a0(980)+/- were defined incorrectly in this PR, which made the test fail. So everything is ok.

@spflueger:  ok, I see.  Your inital comment mentioned D0 --> K0 K+ K-, so I was a
bit puzzled.  Nevertheless, the isospin projection of the a0+ should be
+1, just as for the pi+ (because it has identical quark contents).

If that makes the tests fail, then something else has gone awry.

expertsystem/particle_list.yml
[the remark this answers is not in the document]
redeboer: This PR is derived from the ongoing work for #190. What happens there eventually is that all entries loaded from the external `particle` package should result in the same values as the ones we have in our `particle_list.yml`. That's why this PR is blocking that one: we need these values to be consistent.
As a side note: the `particle` package stores quark content, not isospin projection, so @Leongrim implemented a conversion from that quark content string to isospin. Might be that it function is buggy for some of those quark content strings.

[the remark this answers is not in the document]
redeboer: It's this function
https://github.com/ComPWA/expertsystem/blob/6dd3d8a34ff69678c6b8339a67715d1d682aa26d/expertsystem/io/_pdg.py#L76-L99
Probably doesn't work for superpositions of spin-1 states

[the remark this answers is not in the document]
spflueger: Ok, i will comment in PR #190 on this.

spflueger:
> > hmmm ... consider the decay D0 --> a0(980)+ K- on the quark level: [c ubar] --> [s ubar] W+ --> [s ubar] [u dbar] ~ K- a0+ the a0+ then decays like this: [u dbar] --> [u (sbar s) dbar] --> [u sbar] [s dbar] ~ K+ K0_bar so, the correct decay to consider would be D0 --> K0_bar K+ K-. No? @wgradl Maybe there is slight misunderstanding. We are currently testing `D0 --> K0_bar K+ K-`. Just the isospin of a0(980)+/- were defined incorrectly in this PR, which made the test fail. So everything is ok.
> @spflueger: ok, I see. Your inital comment mentioned D0 --> K0 K+ K-, so I was a bit puzzled. Nevertheless, the isospin projection of the a0+ should be +1, just as for the pi+ (because it has identical quark contents). If that makes the tests fail, then something else has gone awry.

Ah I see, very valuable comment. Currently the test just checks if this is generally possible. Since it will also try pure weak interactions (for both vertices/nodes) it will also work with a K0. Hence I used the terms `K0` and `K0_bar` abit sloppy here. 

I conclude two important points from your comment:

* we should make the second decay vertex strong interaction only. This would actually forbid a `K0`.
* we can think about implementing CKM matrix (roughly) to further rate different decay channels based on their probability or strength

expertsystem/particle_list.yml
```diff
@@ -280,28 +291,30 @@ ParticleList:
...
   a0(980)+:
     PID: 9000211
-    Mass: 0.994
+    Mass: 0.98
+    Width: 0.075
     QuantumNumbers:
       Spin: 0
       Charge: 1
       Parity: +1
-      IsoSpin: { Value: 1, Projection: 1 }
```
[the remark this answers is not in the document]
redeboer: 6e5ff356b326297b72f04d1954523c7fb5a9d841

expertsystem/data.py
```diff
@@ -215,6 +216,37 @@ def __repr__(self) -> str:
         return f"{self.__class__.__name__}{self.name, self.pid, self.state, self.mass, self.width}"
 
 
+def compute_gellmann_nishijima(state: QuantumState) -> Optional[float]:
```
spflueger, line 235: Better change this to return a `int`

tests/data/test_particle.py
```diff
@@ -60,3 +61,24 @@ def test_particle():
...
+
+
+@pytest.mark.parametrize(
+    "state, expected",
+    [
+        (
+            QuantumState(
+                spin=0.0, charge=1, isospin=Spin(1.0, 0.0), strangeness=2,
+            ),
+            1,
```
spflueger, line 73: Maybe add one test with a baryon number as well

tests/data/test_particle.py
```diff
@@ -9,21 +9,24 @@
     compute_gellmann_nishijima,
```
spflueger, line 9: Its a bit confusing that we have two, `test_particle.py` files. Would be good to merge them

redeboer: Eventually, yes. But this is related to #197. For now the `tests` folders mirror the structure of the submodules.

expertsystem/data.py
redeboer: I did that at first, but then figured this is implicit behaviour. Afaik formula itself holds just as well for `float` charges.

spflueger: But the charge from the `QuantumState` or your tests are using `int`. So then a potential cast to int will be done at that point of comparison. Hence I would rather put it in that function. It might hold for float charges, but actually measurable particle has float charge. So you could also filter this problem there. `if (charge).is_integer(): return int(charge)`

redeboer: I think the filtering should take place here:
https://github.com/ComPWA/expertsystem/blob/c0abac05fc3d50bb638f079dc77bd600e1d3e227/expertsystem/data.py#L192-L195
This check will fail if the GM fomula returned a `float` that is not an integer.

tests/data/test_particle.py
redeboer: 83167cd1f60d5fa293f3eded2b73c982d0eb9dfe

expertsystem/data.py
redeboer: > `if (charge).is_integer(): return int(charge)`

To be clear: this would also mean raising an exception if the charge is not castable to `int`. For me that is unexpected behaviour in a function that is just supposed to compute the Gell-Mann–Nishijima formula (as its name states).

spflueger: But by definition this should always be an integer. So if that happens to be not an integer you can raise and exception at the point of cause. That would make more sense to me

redeboer: Okay, let me loosen my reasoning here and go the physics route in an attempt to prove that the GM formula can only concern integers (I haven't found this as an explicit statement in any papers yet).
The GM formula relates baryon number, strangeness etc. The fact that we talk about baryon number means that we cannot have isolated quarks (right, or is there some NP that I'm overlooking here?). That means that the `QuantumState` we are testing here can only have whole charge.
It still seems like a lot of implicit assumptions for a function that should just compute a formula, but I'll put that intelligence in there.

spflueger: Hmm, you have a point. So I think you are right, we do not have to put any more assumptions in there. I just want to make sure we are actually catching out the ill defined particles.

spflueger: I think because `particle.charge` is of type `int`, we are on the safe side.

redeboer: `Particle` instances with `float` charge raise an exception, even if they comply with the Gell-Mann–Nishijima formula. `QuantumState` instances are allowed, but there's no check on GM formula there anyway.

spflueger: And mypy would fish out float charges anyways right?

redeboer: Yes, but mypy isn't checked on outside scripts, it's only used as a consistency check for the framework and tests themselves. But for a user running, say, a Jupyter notebook, an exception would be raised for any `Particle` that has unphysical charge (as per Gell-Mann–Nishijima formula + cast check). `QuantumState` remains an internal matter, however.

PR #203, opened 2020-08-13
refactor: use tilde for antiparticles

redeboer:
In accordance with the [scikit-hep `particle` package](https://github.com/scikit-hep/particle)


tests/state/test_particle.py
```diff
@@ -48,3 +48,17 @@ def test_create_antiparticle(particle_name, anti_particle_name):
...
+def test_create_antiparticle_tilde():
+    anti_particles = DATABASE.find_subset("~")
+    assert len(anti_particles) == 9
+    for anti_particle in anti_particles.values():
+        particle_name = anti_particle.name.replace("~", "")
+        particle = DATABASE[particle_name]
+        anti_particle_name = particle_name + "~"
+        anti_particle_name = anti_particle_name.replace("0~", "~0")
+        created_anti_particle = create_antiparticle(
+            particle, anti_particle_name
```
spflueger: You could also just create the particle from antiparticle directly and compare

tests/amplitude/test_yaml_helicity.py
```diff
@@ -19,7 +19,8 @@
 SCRIPT_PATH = dirname(realpath(__file__))
 
 
-def create_amplitude_generator():
+@pytest.fixture(scope="module")
+def amplitude_generator():
```
spflueger: why not rebase?

tests/amplitude/test_yaml_canonical.py
```diff
@@ -49,44 +49,42 @@ def equalize_dict(input_dict):
     return output_dict
 
 
-class TestCanonicalAmplitudeGeneratorYAML:
```
spflueger: Rebase

redeboer:
For this PR, I did a merge instead of a rebase to keep the original commits available, as these commits were referenced in other PRs and issues (even in scikit-hep/particle). The merge was quite painful (just as a rebase would have been), so I'm not keen on going through that again.

tests/amplitude/test_yaml_helicity.py
redeboer: https://github.com/ComPWA/expertsystem/pull/203#issuecomment-674417427

tests/state/test_particle.py
redeboer: No, because then the names would have been unequal. The `create_antiparticle` function prepends `"anti-"` if you don't provide a name.

spflueger: cant you provide particle_name as a name?

spflueger:
> For this PR, I did a merge instead of a rebase to keep the original commits available, as these commits were referenced in other PRs and issues (even in scikit-hep/particle). The merge was quite painful (just as a rebase would have been), so I'm not keen on going through that again.

I see. Do you have an example of such a commit reference? How to avoid such things in the future? I would say we dont have two PRs which work on the same thing open at once (or one in draft mode).

redeboer:
https://github.com/scikit-hep/particle/pull/251#issuecomment-673374234

redeboer:
I did a rebase and a force push, see whether that causes problems with the existing review and links to commits. I also only left in the part where anti-particles get a `~` instead of `bar`.

spflueger:
> > Do you have an example of such a commit reference?
> 
> [scikit-hep/particle#251 (comment)](https://github.com/scikit-hep/particle/pull/251#issuecomment-673374234)
> 
> > How to avoid such things in the future?
> 
> I don't really understand the problem.

Let me be blunt, It's just a pain reviewing this. Since those merge commits appear, obviously there were merge conflicts. So I have to review the whole merge commit code, as some conflict merging mistakes could have been made. In a rebase I would really see only the changes that were actually made in this PR.

As a side note, I'm really curious how github does the squash merging then. Because I believe that rebase would not work in this case (with the merge commits sowhere in the middle)

spflueger:
> I did a rebase and a force push, see whether that causes problems with the existing review and links to commits. I also only left in the part where anti-particles get a `~` instead of `bar`.

Yeah, as far as I know it will just point to some stale commits. I think its better to refer to a PR or issue, rather than commits.

redeboer:
> I think its better to refer to a PR or issue, rather than commits.

It depends. I guess that if PRs remain small, it's not necessary to refer to individual commits. But in GitHub a nice feature is that permalinks to lines render as snippets, so I use that often. Then the commits need to exist, of course.

spflueger:
> > I think its better to refer to a PR or issue, rather than commits.
> 
> It depends. I guess that if PRs remain small, it's not necessary to refer to individual commits. But in GitHub a nice feature is that permalinks to lines render as snippets, so I use that often. Then the commits need to exist, of course.

Using a commit ref in a PR is fine, but anywhere else I dont think it makes much sense

redeboer:
I think it's possible to enforce rebasing in a PR.

tests/unit/state/test_particle.py
```diff
@@ -58,3 +58,15 @@ def test_create_antiparticle(
...
+def test_create_antiparticle_tilde():
+    anti_particles = DATABASE.find_subset("~")
+    assert len(anti_particles) == 9
+    for anti_particle in anti_particles.values():
+        particle_name = anti_particle.name.replace("~", "")
+        particle = DATABASE[particle_name]
+        created_anti_particle = create_antiparticle(
+            particle, anti_particle.name
+        )
+        assert created_anti_particle == anti_particle
```
spflueger, line 72: I meant this

```python
particle_name = anti_particle.name.replace("~", "")
created_anti_particle = create_antiparticle(
    anti_particle, particle_name
)
assert created_anti_particle == DATABASE[particle_name]
```

redeboer: Ok b323285ae9e896843a956e810be72120c54b9661

PR #211, opened 2020-08-14
test!: separate channel and unit tests

### LINKED ISSUE #210: Separate channel tests from unit tests
Would be nice to separate the unit tests from the integration/system tests. Proposed structure of the [`tests`](https://github.com/ComPWA/expertsystem/tree/2999d2d6130be23d199ef13e626fae216d2fac4f/tests) folder:

- `test`
  - `channel`
    - `test_d0_to_ks_kp_km.py`
    - `test_jpsi_to_gamma_pi0_pi0.py`
    - etc...
  - `unit`
    - `amptitude`
    - `io`
    - etc...

Some thoughts:
- Not sure what to do with [`test_nbody_reactions.py`](https://github.com/ComPWA/expertsystem/blob/2999d2d6130be23d199ef13e626fae216d2fac4f/tests/channels/test_nbody_reactions.py)...
- Keep the names short. That way you can for instance quickly run `pytest test/unit`
- In accordance with `doc`, we call the tests folder `test`
- Let the `tox -e py` test simply run `pytest test/unit` (not this `-m "not slow"` stuff)
- Let the `tox -e test` test run `pytest test`
- Let VSCode only scan the `test/unit` folder. That way you can quickly run all **unit** tests from there.

Better would be to change the channel tests into Jupyter notebooks (after all, they are run and tested as well) and present them in a nice way on RTD. But for that it is necessary to write more unit tests and increase test coverage.

redeboer:
Closes #210 

Makes it easier to run the fast unit tests only. With some additional optimisations, these tests now run in around 20 seconds.


tests/unit/amplitude/expected_recipe.yml
```diff
@@ -1264,28 +1264,6 @@ ParticleList:
...
-    PID: 9050225
-    Mass: 1.936
-    Width: 0.464
-    QuantumNumbers:
-      Spin: 2
-      Charge: 0
-      Parity: 1
-      CParity: 1
-      GParity: 1
-      IsoSpin: 0
```
spflueger, line 1288: Why was this removed?

.vscode/settings.json
```diff
@@ -30,7 +30,7 @@
   "python.linting.pylintCategorySeverity.refactor": "Information",
   "python.linting.pylintEnabled": true,
   "python.linting.pylintUseMinimalCheckers": false,
-  "python.testing.pytestArgs": ["test", "--no-cov"],
+  "python.testing.pytestArgs": ["test/unit", "--no-cov"],
```
spflueger: I would prefer to have them all in there, you can still select the group of tests to run. Or whats your reasoning behind this?

test/unit/amplitude/test_yaml_helicity.py
```diff
@@ -21,10 +21,11 @@
...
 def amplitude_generator():
-    initial_state = [("J/psi", [-1, 1])]
-    final_state = [("gamma", [-1, 1]), ("pi0", [0]), ("pi0", [0])]
-
-    stm = StateTransitionManager(initial_state, final_state, ["f"])
+    stm = StateTransitionManager(
+        initial_state=[("J/psi", [-1, 1])],
+        final_state=["gamma", "pi0", "pi0"],
+        allowed_intermediate_particles=["f"],
+    )
```
spflueger: Can you avoid making such "unrelated" changes in the PR. I know its boy scout rule, but it just confuses git. When files are being moved and one line is edited, then it marks it as deleted and creates a new file.

test/unit/amplitude/test_yaml_canonical.py
```diff
@@ -49,44 +49,42 @@ def equalize_dict(input_dict):
     return output_dict
 
 
-class TestCanonicalAmplitudeGeneratorYAML:
```
spflueger: I was also thinking about removing the class, but I thought I leave it in since the purpose would be to structure tests into categories. Not sure if that makes sense (if we will have more tests coming up here).

Again I would prefer to keep the responsibility of one PR cleaner.

.vscode/settings.json
redeboer: Mainly so that you can run all tests at once. Unfortunately, VSCode cannot yet run specific folders.

tests/unit/amplitude/expected_recipe.yml
redeboer: Because I set the allowed intermediate states to `"f0"` as to speed it up. The test only checks whether the recipe file looks as expected.

test/unit/amplitude/test_yaml_helicity.py
redeboer: Review commit by commit

test/unit/amplitude/test_yaml_canonical.py
redeboer: The test file name already serves to bundle the tests related to "canonical amplitude generation in YAML"

.vscode/settings.json
spflueger: We just need some easy thing to also be able to debug non unit tests

test/unit/amplitude/test_yaml_helicity.py
spflueger: I did. I just dont like the way github shows moved files as deleted and new files

redeboer: It went wrong for the `__init__.py` files because there the word `tests` changed into `test`. But I'll  keep it in mind for next time.

.vscode/settings.json
redeboer: 5ca7388fefe03b4e7804b6aabc4e65dc4bf7aa34
Only thing that bothers me is that it's not possible anymore to run only the tests in the `unit` folder. But that's a shortcoming of VSCode.
image

tests/unit/amplitude/expected_recipe.yml
spflueger: Yeah I thought so, but then I dont understand why only this part is removed and not the amplitude parts as well?

redeboer: That is... a good point :thinking:

.vscode/settings.json
spflueger: yeah

spflueger: https://github.com/microsoft/vscode-python/issues/7188 looks like this wont come any time soon. 
I think we just have to get more unit tests and be able to get rid of those channel tests...

redeboer: https://github.com/ComPWA/expertsystem/blob/1d682f52b530d1a75a62a189e642f3ed0261847f/tests/unit/amplitude/test_yaml_helicity.py#L146-L148
Seems that the test is just incomplete. But now I remember this test sort of stalled while #22 is in the making. We didn't want to put too much effort into this temporary `_yaml_adapter` thingy until then.

tests/unit/amplitude/expected_recipe.yml
redeboer: https://github.com/ComPWA/expertsystem/blob/1d682f52b530d1a75a62a189e642f3ed0261847f/tests/unit/amplitude/test_yaml_helicity.py#L146-L148
Seems that the test is just incomplete. But now I remember this test sort of stalled while #22 is in the making. We didn't want to put too much effort into this temporary `_yaml_adapter` thingy until then.

.vscode/settings.json
spflueger: For me it would be enough if I can debug those channel files as a file, so plugging in `if __name__ == "__main__":`
Then we can remove those channel tests from vscode and all is good

redeboer: Yes, I think it would be cool to have the (or some) channel tests as notebooks.

tests/unit/amplitude/expected_recipe.yml
spflueger: I think the problem is that this test is not really a unit test. And this change is just a workaround to make it faster

.vscode/settings.json
redeboer: > For me it would be enough if I can debug those channel files as a file, so plugging in if __name__ == "__main__":

Oh interesting, seems that they already have that. Only not the nbody tests, because they make use of `pytest.mark.parametrize`...

tests/unit/amplitude/expected_recipe.yml
redeboer: That's true. And that is in turn related to the fact that we have no representation of an amplitude module. For now all that test does is write the file out and see if that works (plus a check on the format of the `ParticleList` and `Kinematics` section.

.vscode/settings.json
spflueger: I guess I would just leave all tests in there then. And if you want to run only unit tests you could use the terminal right. That would be an ok tradeoff for now

tests/unit/amplitude/expected_recipe.yml
redeboer: 181d98ebdc87497857f61c34062d246d671260e5
Well, super dirty hack, but I'm not sure how to make this into a proper unit test right now

.vscode/settings.json
redeboer: Personally I want to "run all tests" with a quick shortkey in VSCode without having to open a terminal for that.

PR #215, opened 2020-08-15
test: verify YAML parameters and dynamics

redeboer:
Realised after https://github.com/ComPWA/expertsystem/pull/211#discussion_r470691256 that at least the `Parameters` and `Dynamics` sections can be checked as well, because of #112. Not sure why the `Intensity` section is still slightly different though.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,3 @@
-; This tox document was made to facilitate the local testing workflow.
-; Dependencies are undefined on purpose, because it is assumed that you
-; installed this repository in 'developer mode'!
-
 [tox]
 passenv = PYTHONPATH
 skip_install = True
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -64,7 +60,7 @@ commands =
     pytest {posargs:tests} \
         --cov-fail-under=85 \
         --cov-report=html \
-        --cov-report=xml \
+        --cov-report=xml
```
spflueger, line 63: are those \ needed then?

redeboer: Strange, I don't understand how that line change ended up there :thinking: But yes, that `\` should be necessary

redeboer: For the record, this happens without the `\`:
```
test run-test: commands[1] | --cov=expertsystem
ERROR: InvocationError for command could not find executable cov=expertsystem
```

Issue #196, opened 2020-08-11
Reconsider submodule structure


redeboer:
As the `expertsystem` is slowly refactored and new features are introduced, we need to reconsider the current module structure. Probably best to have a few main modules like `particles`, `amplitude`, and `solving` that signal the main compontents/responsibilities of the package.

Related to #67 

## Proposed structure

Roughly by order of usage (from input to output):

`expertsystem`
`expertsystem.io` (brings together `particles` and `pwa`)
`expertsystem.particles` ('input')
`expertsystem.reaction` (?)
`expertsystem.reaction.topology` (includes both graph and topology?)
`expertsystem.reaction.propagation`
`expertsystem.reaction.rules` (includes both `conservation_rules` and `solvers`?)
`expertsystem.pwa` ('output')

Deeper submodules can be created, but are isolated implementation (not called by the other modules!) and are can be made private.

spflueger:
Sounds good. Can you make a new proposal package structure so we can fine tune this?

redeboer:
> Can you make a new proposal package structure so we can fine tune this?

Yeah let's keep track of that in the description.

PR #216, opened 2020-08-16
feat: add topology.graph.DotGenerator for visualization of the topology

jiangyi15:
Simple class for 
>
> - **State transition graph**
>  - [ ] Visualization of the topology ([`graphviz`](https://pypi.org/project/graphviz/))


redeboer:
I've done some quick searches for Python alternatives to Graphviz, but it seems they all expect a system installation of graphviz. I'm not sure about what matplotlib can provide in terms of (networklike) graphs. Maybe @Leongrim and @sebastianJaeger know more about that library?

jiangyi15:
The main parts of graphviz is the layout algorithm, so most package need a system installation of graphviz. So, in my class, i just write a dot file, that is a easy way to avoid dependence of graphviz. There also some package can draw such diagram, such as [pyfeyn](https://pypi.org/project/pyfeyn/).

spflueger:
> The main parts of graphviz is the layout algorithm, so most package need a system installation of graphviz. So, in my class, i just write a dot file, that is a easy way to avoid dependence of graphviz. There also some package can draw such diagram, such as [pyfeyn](https://pypi.org/project/pyfeyn/), [networkx](https://pypi.org/project/networkx/).

Good point. I'm also not aware of how "standard" the dot graph language is. Do you think it will easily be possible to create those graph pictures directly from python, like when using jupyter? 
My current research showed that there are not many things available for drawing of "feynman" related graphs (edges connected to only a single node). So I'm not sure drawing it through networkx for example will work. Therefore, if we want to avoid users having to install `graphviz` separately, it seems like we have to write some custom plotting code for creating plots for example via matplotlib etc.

Maybe we can add this plotting functionality as a second step. So a converter from dot language graph to a picture created by matplotlib or so. And for now users could use graphviz.

Not sure what others think about this?

jiangyi15:
Converting dot to graph pictures need the `layout` algorithm, that is what graphviz provided. So, in jupyter, 
```
graph = graphviz.Source(dot_source) 
graph
```
where show the graph easily.

Networkx is a general package for network, we can use  `networkx.draw` to  draw the graph on matplotlib. For feynman diagram, we can add external point of initial and final states, it is the same as only one point (that is the same as graphviz).

redeboer:
> So, in my class, i just write a dot file, that is a easy way to avoid dependence of graphviz.

Ah yes, that's a smart implementation actually. We will have to test this functionality though (note the fact [that the patch test coverage fails](https://github.com/ComPWA/expertsystem/pull/216#issuecomment-674487770)).

The question is, how to test whether the output of `DotGenerator` results in valid dot language, without forcing the user/developer to install Graphviz system-wide. I would say two options (possibly both):

1. Find a Python-only package that can verify dot language and write a unit test with that. This will solve the coverage problem.
2. Write a new Jupyter notebook that is (just as the existing [`quickstart.ipynb`](https://github.com/ComPWA/expertsystem/blob/02265cfe1b3a06a42a3a3405c8c42dc546be4eb4/examples/quickstart.ipynb) and [`particles.ipynb`](https://github.com/ComPWA/expertsystem/blob/02265cfe1b3a06a42a3a3405c8c42dc546be4eb4/examples/particles.ipynb)) tested in our CI (there, we do have the control to install graphviz). This has the additional benefit that the notebook illustrates how the `StateTransitionGraph`s will be rendered on the [usage pages](https://pwa.readthedocs.io/projects/expertsystem/en/0.3.0-alpha/usage.html) (_assuming Read the Docs provides a system-wide install of graphviz!_) and that we can see how class works.
  Note that Jupyter notebook can call `pip install` and (I think) even check for the existence of Graphviz on the system, so we don't need the user to install Graphviz there (the notebook will just emit an error).

jiangyi15:
> Find a Python-only (!) package that can verify dot language and write a unit test with that. This will solve the coverage problem.
[pydot](https://pypi.org/project/pydot/) can do that.

redeboer:
> `pydot` can do that.

It's still not what I was hoping for, because even `pydot` still requires a system-wide installation of Graphviz (so independent of Python):

_"GraphViz: (...) Should be installed separately, using your system's package manager, something similar (e.g., MacPorts), or from its source."_
https://github.com/pydot/pydot#dependencies

This would mean that the tests will fail on some systems and that our [`test` requirements](https://github.com/ComPWA/expertsystem/blob/02265cfe1b3a06a42a3a3405c8c42dc546be4eb4/setup.cfg#L58-L60) are misleading.

jiangyi15:
> > `pydot` can do that.
> 
> It's still not what I was hoping for, because even `pydot` requires a system-wide installation of Graphviz (so independent of Python):
> 
> _"GraphViz: (...) Should be installed separately, using your system's package manager, something similar (e.g., MacPorts), or from its source."_
> https://github.com/pydot/pydot#dependencies
> 
> This would mean that the tests will fail on some systems and that our [`test` requirements](https://github.com/ComPWA/expertsystem/blob/02265cfe1b3a06a42a3a3405c8c42dc546be4eb4/setup.cfg#L58-L60) are misleading.

That dependence is only used for rendering dot to pictures file ([Dot.write](https://github.com/pydot/pydot/blob/master/pydot.py#L1822)), the other parts can work without graphviz. So, it can run on tests.

redeboer:
Okay, cool! Can you add `pydot` to the [`test` requirements](https://github.com/ComPWA/expertsystem/blob/02265cfe1b3a06a42a3a3405c8c42dc546be4eb4/setup.cfg#L58-L60) and write a small test for the `DotGenerator` class? You can get an indication of test coverage with `tox -e py` (assuming you have installed with `pip install -e .[dev]`).

It would also be nice to have a notebook that does render the `DotGenerator` output.
What I'm interested in for now is just some notebook cells that show how to use the `DotGenerator` (and something like the `graphviz` package or `networkx`, up to you) to visualise the `StateTransitionGraph`. I can help with the text later on ;)

tests/channels/test_d0_to_ks_kp_km.py
```diff
@@ -3,8 +3,10 @@
 """
 
 import logging
+import dot_parser
 
 from expertsystem.ui import StateTransitionManager
+from expertsystem.topology.graph import DotGenerator
```
redeboer, line 9: Check import order (`pylint` check). It's best to activate `pre-commit`, see
https://pwa.readthedocs.io/projects/expertsystem/en/0.3.0-alpha/contribute.html#pre-commit

tests/channels/test_d0_to_ks_kp_km.py
```diff
@@ -3,7 +3,7 @@
 """
 
 import logging
-import dot_parser
+import dot_parser  # type: ignore
```
redeboer, line 6: Actually this mypy check can be fixed for the whole framework in the `setup.cfg` file, like this:
https://github.com/ComPWA/expertsystem/blob/02265cfe1b3a06a42a3a3405c8c42dc546be4eb4/setup.cfg#L99-L100
(note the alphabetical order, as in https://github.com/ComPWA/expertsystem/pull/216#discussion_r471613796)

spflueger:
> Nice, that's exactly what I meant! Just some small details.
> @spflueger What do you think about putting this under the channel tests? Probably doesn't matter, we should tackle this with #19 and related test issues anyway.

Yeah would be best to have it as a unit test. So taking some fixed graph instances and validating the dot output for those.

redeboer:
> Yeah would be best to have it as a unit test. So taking some fixed graph instances and validating the dot output for those.

@jiangyi15, just for context, since #211 we're trying to distinguish between actual unit tests and integration tests (for instance the channel tests). Eventually, those `tests/channels` tests will be removed or converted into notebooks, because they are slow.

jiangyi15:
> > Yeah would be best to have it as a unit test. So taking some fixed graph instances and validating the dot output for those.
> 
> @jiangyi15, just for context, since #211 we're trying to distinguish between actual unit tests and integration tests (for instance the channel tests). Eventually, those `tests/channels` tests will be removed or converted into notebooks, because they are slow.

Is there any other way to build such solutions? (with edge_prop and Name)

redeboer:
> Is there any other way to build such solutions? (with edge_prop and Name)
> https://github.com/ComPWA/expertsystem/pull/216#issuecomment-675007219

@spflueger Maybe share a fixture of a recipe, like this one? Then other tests can also be performed on it without affecting speed.
https://github.com/ComPWA/expertsystem/blob/1ebb9eb9bb46570abce3433ca6cd67e98017fbb9/tests/unit/amplitude/test_yaml_helicity.py#L22-L37

redeboer:
> It would also be nice to have a notebook that does render the `DotGenerator` output.
> What I'm interested in for now is just some notebook cells that show how to use the `DotGenerator` (and something like the `graphviz` package or `networkx`, up to you) to visualise the `StateTransitionGraph`. I can help with the text later on ;)
> https://github.com/ComPWA/expertsystem/pull/216#issuecomment-674964244

@jiangyi15 see https://github.com/jiangyi15/expertsystem/pull/1

spflueger:
> > Is there any other way to build such solutions? (with edge_prop and Name)
> > [#216 (comment)](https://github.com/ComPWA/expertsystem/pull/216#issuecomment-675007219)
> 
> @spflueger Maybe share a fixture of a recipe, like this one? Then other tests can also be performed on it without affecting speed. See #220.

Yeah, this is a good idea. Although, keep in mind that this will still be "slow" if its run inside the unit tests. Because the how expertsystem has to generate the solutions at this point.

spflueger:
> > > Yeah would be best to have it as a unit test. So taking some fixed graph instances and validating the dot output for those.
> > 
> > 
> > @jiangyi15, just for context, since #211 we're trying to distinguish between actual unit tests and integration tests (for instance the channel tests). Eventually, those `tests/channels` tests will be removed or converted into notebooks, because they are slow.
> 
> Is there any other way to build such solutions? (with edge_prop and Name)

Unfortunately we do not have an separate graph building functionality except the one in the `StateTransitionManager` (`prepare_graphs()` etc). I was thinking of specifying a "hard coded" graph as a test case. It does not even have to make sense on a physics level.
The idea of @redeboer of using a fixture is also a good solution. Although it would be a bit slower from the runtime (so not perfectly fit into the picture of a unit test)

redeboer:
So this is what we meant by hidden dependencies :(
https://github.com/ComPWA/expertsystem/pull/216/checks?check_run_id=998049035

For the CI, it is indeed solved with 059ddeeccc43ad819b551dafede38dbc6d0933dc (:+1:), but the additional note is important.
https://github.com/ComPWA/expertsystem/blob/d4ff6b10a8c5755553f9568ae079c610470d496e/examples/visualization.ipynb#L106-L107

cspell.json
```diff
@@ -81,6 +81,7 @@
         "toctree",
         "topness",
         "venv",
+        "visualise",
```
redeboer, line 84: Sorry my bad, should be "visualize" in the `vizualisation.ipynb`, because [we use American English](https://pwa.readthedocs.io/projects/expertsystem/en/0.3.0-alpha/contribute.html#spelling). So please remove this line and change to "visualize" in the notebook.

redeboer:
@jiangyi15, thanks for the additions.
I'm currently trying to fix https://github.com/ComPWA/expertsystem/pull/220#issuecomment-675409701, so that a fixture becomes available for the solutions (list of `StateTransitionGraphs`) and the test can be converted into more of a unit test.

Meanwhile, I have a question, also for @spflueger:
Why use a `DotGenerator` class with static methods and not just one function? I'm even considering this functionality should be put [under the `io` module](https://github.com/ComPWA/expertsystem/tree/4f6f6a0287e93286ec67db1dfb40966de276e909/expertsystem/io/__init__.py), with the implementation details (currently class attributes, such as `dot_head` etc) under a private module, like `io._graphviz`.

jiangyi15:
> Why use a `DotGenerator` class with static methods and not just one function? I

In my package, this class will do more things [DotGenerator.dot_chain](https://gitlab.com/jiangyi15/tf-pwa/-/blob/dev/tf_pwa/vis.py#L29), but this method is standalone, so it is a static method.

spflueger:
> > Why use a `DotGenerator` class with static methods and not just one function? I
> 
> In my package, this class will do more things [DotGenerator.dot_chain](https://gitlab.com/jiangyi15/tf-pwa/-/blob/dev/tf_pwa/vis.py#L29), but this method is standalone, so it is a static method.

Correct me if I'm wrong. But it seems like [DotGenerator.dot_chain](https://gitlab.com/jiangyi15/tf-pwa/-/blob/dev/tf_pwa/vis.py#L29) is doing the same for TF-PWA as `from_graph` is doing for the expertsystem. And besides that, there is no functionality in the `DotGenerator`. Could you clarify which other functionalties one could put into the `DotGenerator`, so that a class would be justified?
To me it should be just a function taking as input a graph and outputting a string in the dot format.

redeboer:
> To me it should be just a function taking as input a graph and outputting a string in the dot format.

Note: this is also why I'm tempted to put it into the `io` module. It's similar behaviour as
https://github.com/ComPWA/expertsystem/blob/4f6f6a0287e93286ec67db1dfb40966de276e909/expertsystem/io/__init__.py#L27-L35
although here, the output would be a `str`.

jiangyi15:
> > > Why use a `DotGenerator` class with static methods and not just one function? I
> > 
> > 
> > In my package, this class will do more things [DotGenerator.dot_chain](https://gitlab.com/jiangyi15/tf-pwa/-/blob/dev/tf_pwa/vis.py#L29), but this method is standalone, so it is a static method.
> 
> Correct me if I'm wrong. But it seems like [DotGenerator.dot_chain](https://gitlab.com/jiangyi15/tf-pwa/-/blob/dev/tf_pwa/vis.py#L29) is doing the same for TF-PWA as `from_graph` is doing for the expertsystem. And besides that, there is no functionality in the `DotGenerator`. Could you clarify which other functionalties one could put into the `DotGenerator`, so that a class would be justified?
> To me it should be just a function taking as input a graph and outputting a string in the dot format.

In my package, the purpose of that class is to deal with a `Particle` instead of `DecayChain`, so there may be some other function for the class. But for `expertsystem` there is only a graph now, so it may be meaningless to be a static method.

spflueger:
> > > > Why use a `DotGenerator` class with static methods and not just one function? I
> > > 
> > > 
> > > In my package, this class will do more things [DotGenerator.dot_chain](https://gitlab.com/jiangyi15/tf-pwa/-/blob/dev/tf_pwa/vis.py#L29), but this method is standalone, so it is a static method.
> > 
> > 
> > Correct me if I'm wrong. But it seems like [DotGenerator.dot_chain](https://gitlab.com/jiangyi15/tf-pwa/-/blob/dev/tf_pwa/vis.py#L29) is doing the same for TF-PWA as `from_graph` is doing for the expertsystem. And besides that, there is no functionality in the `DotGenerator`. Could you clarify which other functionalties one could put into the `DotGenerator`, so that a class would be justified?
> > To me it should be just a function taking as input a graph and outputting a string in the dot format.
> 
> In my package, the purpose of that class is to deal with a `Particle` instead of `DecayChain`, so there may be some other function for the class. But for `expertsystem` there is only a graph now, so it may be meaningless to be a static method.

Hmm, I guess `Particle` is used to put more information on an edge or node? If thats the case it seems like this can be handled by another function though. So I don't yet see the need of a class.

So unless there is some reason to use a class, I would suggest to keep it simple and stick with a function.

jiangyi15:
> Hmm, I guess Particle is used to put more information on an edge or node? 

Particle is the initial states which include all topology in my package.

spflueger:
> > Hmm, I guess Particle is used to put more information on an edge or node?
> 
> Particle is the initial states which include all topology in my package.

So then this generates a whole collection of dot graphs or what is the difference to a graph (or DecayChain)?

Basically my question is, what other parts do we need to generate as a dot string except the graph?

jiangyi15:
The amplitude formula can be another part.

spflueger:
> The amplitude formula can be another part.

I'm a bit lost now. What is an "amplitude formula"? I'm thinking of a mathematical formula now. I thought the dot language is only useful for graphs.

jiangyi15:
> > The amplitude formula can be another part.
> 
> I'm a bit lost now. What is an "amplitude formula"? I'm thinking of a mathematical formula now. I thought the dot language is only useful for graphs.

I mean the computation graph of amplitude.

spflueger:
> I mean the computation graph of amplitude.

Ah, I see. But how is the class going to help there? To implement the dot graph for the computation, you need a completely separate separate collection of functions. There is no common state that the class could handle. Then a collection of functions outside will work just fine

What I'm looking for is a change of the interface from `DotGenerator.from_graph(graph)`, to something like `generate_dot(graph)`. The latter is much clearer to me. Since the former `DotGenerator.from_graph(graph)` suggest that a DotGenerator is initialized from a graph so returning a `DotGenerator` instance. I also still do not see any need for a class at this point.

Or at which point do you disagree with me here?

@redeboer What are your thoughts on this.

redeboer:
> @redeboer What are your thoughts on this.

From my side I can only add that I want to contain any input/output functionality in the `io` module (see also #196). In this case, we want to have a facade function that can take an arbitrary object and convert that to a `str` of [DOT language](https://en.wikipedia.org/wiki/DOT_(graph_description_language)). That facade uses `isinstance` to check the type of the object and will call specific converters, in this case something like the `DotGenerator.from_graph` in the case of a `StateTransitionGraph`.

Another facade could then write that output `str` to an actual [`.gv` file](https://marc.info/?l=graphviz-devel&m=129418103126092), but that would just be an additional service.

Compare:
https://github.com/ComPWA/expertsystem/blob/f189fa694af64092c9aad382eeb0ad59fb26aebf/expertsystem/io/yaml/__init__.py#L38-L44

@spflueger If you agree with such a structure, I can make a PR to @jiangyi15 's master branch to illustrate the idea (I also have to think how to organise this specifically within the `io` module).

spflueger:
> > @redeboer What are your thoughts on this.
> 
> From my side I can only add that I want to contain any input/output functionality in the `io` module (see also #196). In this case, we want to have a facade function that can take an arbitrary object and convert that to a `str` of [DOT language](https://graphviz.org/doc/info/lang.html). That facade uses `isinstance` to check the type of the object and will call specific converters, in this case something like the `DotGenerator.from_graph` in the case of a `StateTransitionGraph`.
> 
> Another facade could then write that output `str` to an actual [`.gv` file](https://marc.info/?l=graphviz-devel&m=129418103126092), but that would just be an additional service.

Hmm ok, that sounds reasonable. But what to do about his `DotConverter`. Should it stay a class in your opinion? Also @jiangyi15 please point out to me if I'm wrong and this `DotConverter` should stay a class. :smile:

redeboer:
> Hmm ok, that sounds reasonable.

Will do :white_check_mark: 

> Yeah would be best to have it as a unit test.
> https://github.com/ComPWA/expertsystem/pull/216#issuecomment-675003053

Should be addressed with https://github.com/jiangyi15/expertsystem/pull/2 now

redeboer:
@spflueger https://github.com/jiangyi15/expertsystem/pull/3 illustrates https://github.com/ComPWA/expertsystem/pull/216#issuecomment-675556837

spflueger (approved):

redeboer (approved):

redeboer:
Many thanks, @jiangyi15! :tada:

PR #218, opened 2020-08-17
test: split ParticleCollection tests io/data

redeboer:
* `test_particle_collection.py` now only checks the interface of `ParticleCollection`, while `test_particle_list.py` only checks how `ParticleCollection` whether written and read correctly through the `io` module.
* The `particle.DATABASE` is now available to _all_ unit tests as a fixture.

These tests need to be split before #190, because there, a new input test will be written (input from the PDG).


tests/unit/conftest.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+from expertsystem import ui
+from expertsystem.data import ParticleCollection
+from expertsystem.state.particle import DATABASE
+
+
+@pytest.fixture(scope="module")
```
spflueger, line 8: Interesting, so because pytest is collecting all tests, this will also be loaded automatically right? I guess this will not allow to test `test_particle.py` individually. I assume some kind of import will be needed.

Btw, I think the scope of this can be set to `package`

redeboer: Yes, see
https://docs.pytest.org/en/stable/fixture.html#conftest-py-sharing-fixture-functions

What I find more remarkable is that `mypy`, `pylint`, etc somehow understand this shared fixture without an `import` statement.

> Btw, I think the scope of this can be set to package

See a677a67a9fae11e06c749be76652a4026b1a3a31

PR #220, opened 2020-08-18
refactor: share test solutions with fixture

redeboer:
Can be used by #216

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,11 +1,66 @@
+# pylint: disable=redefined-outer-name
+import logging
+
 import pytest
 
 from expertsystem import ui
+from expertsystem.amplitude.canonical_decay import CanonicalAmplitudeGenerator
+from expertsystem.amplitude.helicity_decay import HelicityAmplitudeGenerator
 from expertsystem.data import ParticleCollection
 from expertsystem.state.particle import DATABASE
+from expertsystem.ui import (
+    InteractionTypes,
+    StateTransitionManager,
+)
+
+
+logging.basicConfig(level=logging.ERROR)
 
 
 @pytest.fixture(scope="package")
 def particle_database() -> ParticleCollection:
     ui.load_default_particle_list()
     return DATABASE
+
+
+@pytest.fixture(scope="module")
+def canonical_solutions():
+    stm = StateTransitionManager(
+        initial_state=[("J/psi(1S)", [-1, 1])],
+        final_state=["gamma", "pi0", "pi0"],
+        allowed_intermediate_particles=["f(0)"],
+        formalism_type="canonical-helicity",
+    )
+    stm.set_allowed_interaction_types([InteractionTypes.EM])
+    graph_interaction_settings_groups = stm.prepare_graphs()
+    solutions, _ = stm.find_solutions(graph_interaction_settings_groups)
+    return solutions
+
+
+@pytest.fixture(scope="module")
+def helicity_solutions():
+    stm = StateTransitionManager(
+        initial_state=[("J/psi(1S)", [-1, 1])],
+        final_state=["gamma", "pi0", "pi0"],
+        allowed_intermediate_particles=["f(0)"],
+    )
+    stm.set_allowed_interaction_types(
+        [InteractionTypes.Strong, InteractionTypes.EM]
+    )
+    graph_interaction_settings_groups = stm.prepare_graphs()
+    solutions, _ = stm.find_solutions(graph_interaction_settings_groups)
+    return solutions
+
+
+@pytest.fixture(scope="module")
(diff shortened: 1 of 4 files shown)


tests/unit/conftest.py
```diff
@@ -1,11 +1,66 @@
...
+    amplitude_generator = CanonicalAmplitudeGenerator()
+    amplitude_generator.generate(canonical_solutions)
+    return amplitude_generator
+
+
+@pytest.fixture(scope="module")
+def helicity_amplitude_generator(helicity_solutions):
+    amplitude_generator = HelicityAmplitudeGenerator()
+    amplitude_generator.generate(helicity_solutions)
+    return amplitude_generator
```
spflueger, line 66: Will those fixtures actually be used in many different tests modules?

redeboer: See https://github.com/ComPWA/expertsystem/pull/216#issuecomment-675007219

spflueger: But those wont be used by the `DotGenerator` test right? Just the solution fixtures

redeboer: The `DotGenerator` needs some solution samples to test on in the unit tests. I guess there are other features of amplitudes/solutions that can be 'unit-test probed' as well.

tests/unit/conftest.py
```diff
@@ -1,11 +1,66 @@
...
 
 
 @pytest.fixture(scope="package")
 def particle_database() -> ParticleCollection:
     ui.load_default_particle_list()
     return DATABASE
+
+
+@pytest.fixture(scope="module")
+def canonical_solutions():
```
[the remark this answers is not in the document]
redeboer: I'm trying to do this now, but the names become a mess:
`jpsi_to_pi_pi_gamma_helicity_solutions`
`jpsi_to_pi_pi_gamma_canonical_amplitude_generator`
etc...

Sure we want to go that route? :grimacing:

tests/unit/conftest.py
spflueger: Excuse me but I need to follow up on this.
So my statement is correct right. Then they are just moved there in case they will be used by other tests. I would prefer in the future to not move things around without any reason. Also we should change the names of those fixtures if they are moved into a global test scope, as they are quite uninformative

redeboer: > I would prefer in the future to not move things around without any reason.

Yeah there's a conflict here between keeping PRs small and keeping modifications coupled. My mistake here was no to link the PR in zenhub as an issue dependency.

> Then they are just moved there in case they will be used by other tests.

Generally, this goes into unit test design, which is a difficult topic and I consider part of the #19 epic.

> Also we should change the names of those fixtures if they are moved into a global test scope, as they are quite uninformative

Maybe call them `dummy_helicity_solutions` etc? I just want to avoid names that become like 79 characters wide.

spflueger: would `jpsi_pipigamma_helicity_solutions` be too long? Dunno, I guess im also fine with dummy.

redeboer: Okay. Only thing is that `pipigamma` is not acceptable to `cspell`.
e5035a7eb6e3e676a44ac80a27db4af172bead2c

tests/unit/conftest.py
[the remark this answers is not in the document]
spflueger: Well in my view the generators do not belong here. And if some test uses `canonical_solutions`, what to expect there? Before they were bundled with the test module itsself, so it is a bit more encapsulated.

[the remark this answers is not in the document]
spflueger: I'm not saying we have to do this. I'm merely stating my opinion here

tests/unit/conftest.py
spflueger: Hmm would be ok to me. Or are those names too displeasing?

redeboer:
I don't understand why RTD is suddenly protesting... :roll_eyes: 
https://readthedocs.org/projects/expertsystem/builds/11686583/
We fixed all the requirements, right?
https://github.com/ComPWA/expertsystem/blob/8132ac5c33107464b197341f9a080facf35b3858/setup.cfg#L61-L68

redeboer:
Same here...
https://readthedocs.org/projects/tensorwaves/builds/11686495/
Looks like RTD did some wacky upgrade between 02b9b50f48b75c8d4c7e41e28f8235f1279f8f96 and e5035a7eb6e3e676a44ac80a27db4af172bead2c

tests/unit/conftest.py
redeboer: It's just the way `cspell` works. If we want to allow `pipigamma`, it should be added as an allowed word, but I think that's an arbitrary choice.

PR #224, opened 2020-08-20
feat!: switch to PDG as source of truth

### LINKED ISSUE #189: Switch to PDG as source of truth

redeboer:
Closes #189


redeboer:
@spflueger 
- I left the `particle_list.*ml` files in. It will be addressed in a follow-up PR for #33 
- The tests are still slower than previously, despite the `limited_allowed_intermediate_states` fixture. I think this is because of the way `allowed_intermediate_states` works (partial `str` search)
https://github.com/ComPWA/expertsystem/blob/df9f9e06f3ac51f593b48bdfca26e346868a5dfd/tests/conftest.py#L8

redeboer:
Also documentation building takes ages now :(

Issue #223, opened 2020-08-19
Isolate development to dev branch


redeboer:
We should move towards a safer branching model: a stable `master` branch as the default branch, plus a `dev` branch for the ongoing development.

Such a model can also facilitate splitting tests into resource intensive integration tests (perhaps also benchmarks https://github.com/ComPWA/tensorwaves/issues/103) that are run only when attempting to merge/rebasr `dev` into `master`, and fast unit tests that are only run on feature branch PRs. (#19)

We also have to think carefully about how best to accommodate the demands of fast research results in academia while not affecting the eventual design.

The same applies to `tensorwaves` but the `expertsystem` is a natural starting point.

See also
- https://git-scm.com/book/en/v2/Distributed-Git-Maintaining-a-Project
- https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow
- https://nvie.com/posts/a-successful-git-branching-model/

redeboer:
Closed by #360

PR #227, opened 2020-08-21
fix!: remove faulty print statement

redeboer:
Oops...
image

https://pwa.readthedocs.io/projects/expertsystem/en/latest/usage/quickstart.html


redeboer:
Can happen, see also #223

Issue #143, opened 2020-07-31
Implement Particle in _create_seed_graphs


redeboer:
_From #172_
Instead of working with the nested dicts internally use a `tuple` of a `Particle` with a `float` as spin projection as the edge properties.

The `__eq__` operation between those classes (as compared to the nested dicts) can also be used to remove or simplify `CompareGraphElementPropertiesFunctor`.

redeboer:
The crucial part that this issue is about it this
https://github.com/ComPWA/expertsystem/blob/0d6da1cad3a4653db64b498d93a48c050396007b/expertsystem/state/particle.py#L575-L579

PR #229, opened 2020-08-24
refactor!: merge topology submodules

redeboer:
This is a spin-off maintenance PR from the work on #143, where I try to distinguish the different `StateTransitionGraphs` (empty topology, filled with `dict`s and filled with `ComplexEnergyState`/`QuantumState`) through generic types.

Major change is that the functions that modify `StateTransitionGraphs` have become methods, which simplifies import statements in several tests, notebooks, and modules that depend on the `topology` submodules.


expertsystem/topology.py
```diff
@@ -0,0 +1,453 @@
+"""All modules related to topology building.
```
spflueger, line 1: Hmm, I think its better to rename this into graph. In my definition a topology would be just a graph without any properties attached to it. A general graph contains properties though, so graph would be the more general term.

expertsystem/ui/_system_control.py
```diff
@@ -465,14 +460,14 @@ def match_external_edges(graphs: List[StateTransitionGraph]) -> None:
     if not graphs:
         return
     ref_graph_id = 0
-    _match_external_edge_ids(graphs, ref_graph_id, get_final_state_edges)
-    _match_external_edge_ids(graphs, ref_graph_id, get_initial_state_edges)
+    _match_external_edge_ids(graphs, ref_graph_id, "get_final_state_edges")
+    _match_external_edge_ids(graphs, ref_graph_id, "get_initial_state_edges")
```
spflueger, line 464: Those string changes are super confusing. I assume that change was made in order to handle both normal function and instance methods of the graph. This would require some improvements

redeboer: You're right, this remains the major problem with embedding the functions as methods: you cannot pass around a method as a function argument. Afaics there's no way around this in Python.

So to summarise: having functions like `get_final_state_edges` as a member of `StateTransitionGraph` is safer and easier in usage (keep responsibilities nearby each other), but (_within the current framework!_) the price is this need for `getattr`. 
I hate that too, because these strings are against static typing and things like tab completion.

redeboer: > I assume that change was made in order to handle both normal function and instance methods of the graph.

The thing is that I didn't want to touch the rest of the framework other than `import` statements. In this case, there are a few functions under `ui` that could become part of the `graph` module (take this `_match_external_edge_ids` for example), but that requires additional refactoring that goes beyond what this PR intended.

spflueger: The problem is that, this rather complicated change (which started of with a simple move functions to method idea) was not discussed and planned in prior. I have to at least think about this now, and see if there is some quick and nice solution for this.

expertsystem/topology.py
redeboer: In a way, this touches on #196. For now I just followed the existing hierarchy.

expertsystem/ui/_system_control.py
redeboer: That's true. Although on the other hand, we probably would have concluded that issue with "sure, sounds good, let's convert the functions to methods", because the consequence of that only pop up after the refactoring itself (this PR). In that sense, issues and PRs cannot be strictly separated.

So I decoupled this PR from #230 (no rebasing) and #143, then we can consider this PR a proposal that can incite other discussions (such as https://github.com/ComPWA/expertsystem/pull/229#discussion_r475536307).

spflueger: Ok, so checked out if there is a nice way with limited amount of changes to solve this. Unfortunately the amount of changes would go indeed beyond the scope of the PR. So we tackle this in another one

expertsystem/topology.py
spflueger: Lets also move that into a separate issue

spflueger (approved):

Issue #231, opened 2020-08-24
Improve graph initialization combinatorics including identical particles


spflueger:
This issue was triggered by PR #229 .

The code handling the combinatorics of graphs with regard to the initial and final state is rather complicated. It is handled in two stages with rather large amounts of code, which is hard to understand code.

These two stages are:
1. `create_seed_graphs`: This handles attaching the initial and final state particles to the external edges including the combinatorics of the spin projections. However it does not include possible identical particle swappings
2. `find_solutions` (in STM): This handles possible swappings of initial or final state edges that are identical particles.

Note: identical particles can be swapped and yield different results under the circumstance that the sets of unique ids of initial/final state ids on some node differs. Consider the following `JPsi -> gamma pi0_1 pi0_2` example:

The decay via f intermediate state `JPsi -> gamma f; f -> pi0_1 pi0_2` is kinematically identical to the one in which the `pi0`s are swapped, so `f -> pi0_2 pi0_1`. Hence the swapping has no effect and can be ignored.
However in the other subsystem `JPsi -> pi0_1 omega; omega -> gamma pi0_2` the swapping of the `pi0`s yields a different kinematic outcome (`JPsi -> pi0_2 omega; omega -> gamma pi0_1`). Therefore this identical particle swapping has to be considered and created as a new graph.

**Possible solutions:**

1. Instead of splitting the task into two, `create_seed_graphs` could also include identical particle combinatorics and remove the current second stage completely.
Also, this means that the **responsibility** of `create_seed_graphs` **changes**! It would take care of identical particles as well.

2. Keep things separated, but try to find a cleaner way of handling this (not sure on the details of this though).

I see a bit of premature optimization here and solution 1. would be conceptually be easier to understand and implement. The downside would be small loss of performance. How much this actually is would have to be tested. Therefore I favor solution 1

Also note that issue #143 is influenced by this, as the responsibility of `create_seed_graphs` could change.

**EDIT**: We decided to go with **option 2**! The identical particle symmetrization part is now done in the amplitude generation (see #243). Which has several advantages:
* Definition of a solution is more straight forward (no identical particle permutations involved)
* Allows definition of normalization factors for amplitudes with identical particle symmetries [e.g.: `1/2 * (A(pi0_1, pi0_2) + A(pi0_2, pi0_1))`]
* Still maintains the performance

spflueger:
I opened the tiny issue #232 which relates to this one

PR #233, opened 2020-08-25
test: reorganize J/psi to gamma pi0 pi0

### LINKED ISSUE #232: Improve readability of JPsi to gamma pi0 pi0 test
The total number of found solutions is 48. It would be good to do a little bit more fined grained testing of the solutions. This would improve the quality of the test and also make it easier to understand.

So the amount of 48 totals up in the following way (note that a total of two f0, two f2 and one omega state are included):
* each f0 intermediate state creates 4 solutions -> 2*4 = **8**
* each f2 intermediate state creates 12 solutions -> 2*12 = **24**
* the omega state creates 8 solutions. **But** identical particle combinatorics increases this by a factor of 2 (swapping the two pi0s). ->  2*8 = **16**

-> 8 + 24 + 16 = 48

I propose to test for the amount of f0, f2 and omega solutions separately

redeboer:


tests/channels/test_jpsi_to_gamma_pi0_pi0.py
```diff
@@ -17,30 +17,54 @@
 logging.getLogger().setLevel(logging.ERROR)
 
 
+@pytest.mark.parametrize(
+    "allowed_intermediate_particles, number_of_solutions",
+    [
+        (["f(0)(1500)"], 4),
+        (["f(0)(980)", "f(0)(1500)"], 8),
+        (["f(2)(1270)"], 12),
+        (["omega(782)"], 16),
```
spflueger, line 26: Can you make a small comment that "basic" number of solutions is 8, but due to identical particle combinatorics it get increased by a factor of 2.

tests/channels/test_jpsi_to_gamma_pi0_pi0.py
```diff
@@ -17,30 +17,54 @@
...
+def test_id_to_particle_mappings():
+    stm = StateTransitionManager(
+        initial_state=[("J/psi(1S)", [-1, 1])],
+        final_state=[("gamma", [-1, 1]), ("pi0", [0]), ("pi0", [0])],
+        allowed_intermediate_particles=["f(0)(980)"],
+        number_of_threads=1,
+    )
+    stm.set_allowed_interaction_types([InteractionTypes.Strong])
+    graph_interaction_settings_groups = stm.prepare_graphs()
+    solutions, _ = stm.find_solutions(graph_interaction_settings_groups)
```
spflueger, line 66: A fixture would be better

spflueger: Ah nvm. I think this test will fall way in the end anyway.

spflueger (approved):
Looks good. Actually I put the dependency on issue #231 in there in purpose, as the `test_id_to_particle_mappings()` will probably change. We can merge that PR, but lets remove the closes statement, so that we take a look at this issue again once #231 is done.

PR #235, opened 2020-08-27
feat: visualize list of StateTransitionGraphs

redeboer:
Previously, DOT visualisation would crash if it got an uninitialised `StateTransitionGraph` (because it doesn't have edge properties). Now, the edge ID is used instead if there are no edge properties. In addition, if there are edge properties, the spin projection is printed as well. A `list` of solutions can now also be rendered in one graphviz file.

A quick demo is available under `test_graph.py`.

Also includes some minor fixes and refactorings to the `particle` submodule.

### DIFF
--- a/expertsystem/io/dot.py
+++ b/expertsystem/io/dot.py
@@ -4,6 +4,8 @@ See :doc:`/usage/visualization` for more info.
 """
 
 from typing import (
+    Any,
+    Callable,
     List,
     Optional,
 )
@@ -44,19 +46,29 @@ _DOT_DEFAULT_EDGE = '    "{}" -> "{}";\n'
 _DOT_LABEL_EDGE = '    "{}" -> "{}" [label="{}"];\n'
 
 
+def embed_dot(func: Callable[[Any], str]) -> Callable[[Any], str]:
+    """Add a DOT head and tail to some DOT content."""
+
+    def wrapper(*args, **kwargs):  # type: ignore
+        dot_source = _DOT_HEAD
+        dot_source += func(*args, **kwargs)
+        dot_source += _DOT_TAIL
+        return dot_source
+
+    return wrapper
+
+
+@embed_dot
 def __graph_list_to_dot(graphs: List[StateTransitionGraph]) -> str:
-    dot_source = _DOT_HEAD
+    dot_source = ""
     for i, graph in enumerate(graphs):
         dot_source += __graph_to_dot_content(graph, prefix=f"g{i}_")
-    dot_source += _DOT_TAIL
     return dot_source
 
 
+@embed_dot
 def __graph_to_dot(graph: StateTransitionGraph) -> str:
-    dot_source = _DOT_HEAD
-    dot_source += __graph_to_dot_content(graph)
-    dot_source += _DOT_TAIL
-    return dot_source
+    return __graph_to_dot_content(graph)
 
 
 def __graph_to_dot_content(
--- a/expertsystem/io/dot.py
+++ b/expertsystem/io/dot.py
@@ -42,7 +42,7 @@ def write(instance: object, filename: str) -> None:
 _DOT_HEAD = """digraph {
     rankdir=LR;
     node [shape=point, width=0];
-    edge [arrowhead=none, labelfloat=true];
+    edge [arrowhead=none];
 """
 _DOT_TAIL = "}\n"
 _DOT_RANK_SAME = "    {{ rank=same {} }};\n"
--- a/expertsystem/io/dot.py
+++ b/expertsystem/io/dot.py
@@ -19,6 +19,15 @@ def convert_to_dot(instance: object) -> str:
     """
     if isinstance(instance, StateTransitionGraph):
         return __graph_to_dot(instance)
+    if isinstance(instance, list):
+        output_dot_string = "digraph {\n"
+        for i, item in enumerate(instance):
+            dot_single_graph = convert_to_dot(item).replace(
+                "digraph", f"subgraph cluster_{i}"
+            )
+            output_dot_string += f"{dot_single_graph}\n"
+        output_dot_string += "}"
+        return output_dot_string
     raise NotImplementedError(
         f"Cannot convert a {instance.__class__.__name__} to DOT language"
     )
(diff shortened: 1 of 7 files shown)


redeboer:
Currently looking whether it's possible to render a list of `StateTransitionGraph` in *one* output file, but this does not have to be included in this PR.

expertsystem/state/particle.py
```diff
@@ -526,27 +529,32 @@ def initialize_graph(
...
+            spin_projections = arange(  # type: ignore
+                -particle.state.spin, particle.state.spin + 1, 1.0
+            ).tolist()
+            if particle.mass == 0.0:
+                if 0.0 in spin_projections:
+                    del spin_projections[spin_projections.index(0.0)]
+            state = (particle_name, spin_projections)
+        return state
+
+    return [safe_set_spin_projections(state) for state in list_of_states]
```
spflueger, line 548: Can you please avoid to do such unrelated refactoring work, if it is being overhauls by upcoming issues anyway

expertsystem/io/dot.py
```diff
@@ -30,55 +32,88 @@ def write(instance: object, filename: str) -> None:
...
-        return ",".join(name_list)
 
+def __graph_to_dot(graph: StateTransitionGraph) -> str:
     dot_source = _DOT_HEAD
+    dot_source += __graph_to_dot_content(graph)
+    dot_source += _DOT_TAIL
+    return dot_source
 
+
+def __graph_to_dot_content(
```
spflueger, line 62: Proposal: change this back to `__graph_to_dot`, but include the prefix part. Remove `__graph_to_dot`

expertsystem/io/dot.py
```diff
@@ -30,55 +32,88 @@ def write(instance: object, filename: str) -> None:
...
 
-def __graph_to_dot(graph: StateTransitionGraph) -> str:
-    def node_name(edge_id: int, node_id: Optional[int] = None) -> str:
-        if node_id is None:
-            return f"edge{edge_id}"
-        return f"node{node_id}"
+def __graph_list_to_dot(graphs: List[StateTransitionGraph]) -> str:
+    dot_source = _DOT_HEAD
+    for i, graph in enumerate(graphs):
+        dot_source += __graph_to_dot_content(graph, prefix=f"g{i}_")
```
spflueger, line 50: just call the proposed `__graph_to_dot` function

expertsystem/io/dot.py
```diff
@@ -30,55 +32,88 @@ def write(instance: object, filename: str) -> None:
...
     outs = graph.get_final_state_edges()
-    for i in top:
-        dot_source += _DOT_DEFAULT_NODE.format(
-            node_name(i), graph.edge_props[i]["Name"]
-        )
-    for i in outs:
+    for edge_id in top + outs:
         dot_source += _DOT_DEFAULT_NODE.format(
-            node_name(i), graph.edge_props[i]["Name"]
+            prefix + __node_name(edge_id), __edge_label(graph, edge_id)
```
spflueger, line 70: why are nodes labeled?

expertsystem/io/dot.py
```diff
@@ -30,55 +32,88 @@ def write(instance: object, filename: str) -> None:
...
-    dot_source += _DOT_RANK_SAME.format(format_particle(outs))
-
+    dot_source += __rank_string(top, prefix)
+    dot_source += __rank_string(outs, prefix)
     for i, edge in graph.edges.items():
         j, k = edge.ending_node_id, edge.originating_node_id
         if j is None or k is None:
             dot_source += _DOT_DEFAULT_EDGE.format(
-                node_name(i, k), node_name(i, j)
+                prefix + __node_name(i, k), prefix + __node_name(i, j)
```
spflueger, line 78: this is confusing. Im asking myself what a node_name has to do with edge labels

spflueger (commented):
I proposed some changes to make it a bit cleaner. I'm not sure about this `node_name` stuff. Is a bit confusing to me.

expertsystem/io/dot.py
redeboer: This comes from #216

expertsystem/state/particle.py
redeboer: Kicked out

expertsystem/io/dot.py
redeboer: I have to think about this. Probably, the construction of wrapping the 'content' into `_DOT_HEAD` and a `_DOT_TAIL` can be better done with a decorator.

expertsystem/io/dot.py
redeboer: That's not how DOT syntax works

expertsystem/io/dot.py
redeboer: For the decorator, see 82ade244b2b8ab8031c870e936a5591dbc212faf
I'm not sure whether combining functions here makes much sense. How would you do it?

spflueger: Well i meant, take this function rename it `__graph_to_dot` and also add the header and footer back in.

expertsystem/io/dot.py
spflueger: Ok, sorry I misunderstood the code then. So to clarify, the head and tail is only required for one single dot file, not for each graph?

expertsystem/io/dot.py
spflueger: I guess I misunderstood this DOT syntax then

expertsystem/io/dot.py
spflueger: yeah ok, but can we easily change that to something meaningful. Maybe I'm missing something here. What I'm saying is, cant we change this into `prefix + str(k), prefix + str(j)`?

spflueger: Ah i think i get it now. You have to create some artificial nodes for the external lines. Ok, so lets keep it like this then and possibly refactor it in another PR.

expertsystem/io/dot.py
redeboer: The problem is that DOT language only works with nodes. So actually the node names are confusing: we prepend `"edge"` to a node that represents the end point of an open edge (initial and final state) and prepend `"node"` to internal nodes. See output of `test_dot.py`:

```dot
digraph {
    rankdir=LR;
    node [shape=point, width=0];
    edge [arrowhead=none];
    "edge0" [shape=none, label="J/psi(1S)[1]"];
    "edge2" [shape=none, label="gamma[1]"];
    "edge3" [shape=none, label="pi0[0]"];
    "edge4" [shape=none, label="pi0[0]"];
    { rank=same "edge0" };
    { rank=same "edge2", "edge3", "edge4" };
    "edge0" -> "node0";
    "node0" -> "node1" [label="f(0)(980)[0]"];
    "node0" -> "edge2";
    "node1" -> "edge3";
    "node1" -> "edge4";
}
```

expertsystem/io/dot.py
redeboer: It's a bit more complicated, but the labeling can indeed be improved https://github.com/ComPWA/expertsystem/pull/235#discussion_r478392499

spflueger (approved):

PR #243, opened 2020-09-08
refactor!: move particle combinatorics

redeboer:
Lead-up to #231: the list of solutions by `StateTransitionGraph.find_solutions` now does not contain the kinematic combinatorics anymore. This responsibility has been moved to the amplitude builders.

### DIFF
--- a/tests/unit/amplitude/test_yaml_canonical.py
+++ b/tests/unit/amplitude/test_yaml_canonical.py
@@ -50,7 +50,7 @@ def test_particle_section(imported_dict):
 
 def test_parameter_section(imported_dict):
     parameter_list = imported_dict["Parameters"]
-    assert len(parameter_list) == 9
+    assert len(parameter_list) == 5
     for parameter in parameter_list:
         assert "Name" in parameter
         assert "Value" in parameter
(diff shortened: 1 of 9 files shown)


spflueger (approved):

tests/unit/amplitude/test_yaml_canonical.py
```diff
@@ -50,7 +50,7 @@ def test_particle_section(imported_dict):
 
 def test_parameter_section(imported_dict):
     parameter_list = imported_dict["Parameters"]
-    assert len(parameter_list) == 9
+    assert len(parameter_list) == 5
```
redeboer, line 53: So does this mean the canonical amplitude always had too many parameters (some of which weren't used)? And, if so, where they used, and therefore always decoupled by mistake?

spflueger: >So does this mean the canonical amplitude always had too many parameters (some of which weren't used)? And, if so, where they used, and therefore always decoupled by mistake?

Only in some cases. A test in `test_y_to_d0_d0bar_pi0_pi0` compared the number of parameters using helicity vs canonical formalism and requires them to be equal. That test has always passed. I'm not sure why this `test_yaml_canonical.py` created 9 instead of 5 parameters though. I simply checked what the helicity counterpart created, which was 5 as well. And the number of parameters should be the same, independent of the underlying formalism

PR #245, opened 2020-09-17
refactor!: improve graph combinatorics

### LINKED ISSUE #143: Implement Particle in _create_seed_graphs

### LINKED ISSUE #231: Improve graph initialization combinatorics including identical particles

redeboer:
Closes #231 
Closes #143 

* #231: Improved readability of combinatorics in `_create_seed_graphs` and `initialize_graphs`
  - Separated kinematic combinatorics from spin combinatorics (two separate steps)
  - Makes final state groupings safer to use through a `KinematicRepresentation` class (implementation may need improvement)

* #143: `StateTransitionGraph` edge properties have become `Tuple(Particle, float)` (particle with spin projection). Currently converted back to `dict`, but this is easy to remove.

_Rebased on #248_


expertsystem/state/particle.py
```diff
@@ -258,6 +264,198 @@ def convert_to_dict(
...
+    initial_state_edge_groups = fill_groupings(
+        get_state_groupings(graph.get_originating_initial_state_edges)
+    )
+    final_state_edge_groups = fill_groupings(
+        get_state_groupings(graph.get_originating_final_state_edges)
+    )
+    return KinematicRepresentation(
+        initial_state=initial_state_edge_groups,
+        final_state=final_state_edge_groups,
+    )
```
spflueger, line 447: I would make this the constructor of the `KinematicalRepresenation` (the docstring can be merged). That is the only use case that is needed and it will not let that cryptic underlying representation (triple nested lists) shine through the class.

expertsystem/state/particle.py
```diff
@@ -464,8 +662,49 @@ def initialize_graph(  # pylint: disable=too-many-locals
...
+) -> List[StateTransitionGraph[ParticleWithSpin]]:
+    def embed_in_list(some_list: List[Any]) -> List[List[Any]]:
+        if not isinstance(some_list[0], list):
+            return [some_list]
+        return some_list
+
+    allowed_kinematic_groupings = None
+    if final_state_groupings is not None:
+        final_state_groupings = embed_in_list(final_state_groupings)
+        final_state_groupings = embed_in_list(final_state_groupings)
```
spflueger, line 677: Does that work if the final state grouping would be of type `List[List[str]]`?

expertsystem/state/particle.py
```diff
@@ -464,8 +662,49 @@ def initialize_graph(  # pylint: disable=too-many-locals
...
+    output_graphs = list()
+    for kinematic_permutation in kinematic_permutation_graphs:
+        spin_permutations = generate_spin_permutations(
+            kinematic_permutation, particles
+        )
+        output_graphs.extend(spin_permutations)
+    return output_graphs
+
+
+def generate_kinematic_permutations(
```
spflueger, line 699: private

expertsystem/state/particle.py
```diff
@@ -540,148 +778,61 @@ def safe_set_spin_projections(
...
-    attached_external_edges_per_node: List[List[int]],
-    edge_particle_mapping: Dict[int, StateWithSpins],
-) -> List[List[StateWithSpins]]:
-    init_edge_lists = []
-    for edge_list in attached_external_edges_per_node:
-        init_edge_lists.append(
-            sorted([edge_particle_mapping[i] for i in edge_list])
-        )
-    return sorted(init_edge_lists)
+def generate_outer_edge_permutations(
```
spflueger, line 781: private

expertsystem/state/particle.py
```diff
@@ -258,6 +264,198 @@ def convert_to_dict(
...
+            return item
+        if isinstance(item, (tuple, list)) and isinstance(item[0], str):
+            return item[0]
+        if isinstance(item, Particle):
+            return item.name
+        if isinstance(item, dict) and "Name" in item:
+            return str(item["Name"])
+        raise ValueError(
+            f"Cannot extract particle name from {item.__class__.__name__}"
+        )
```
spflueger, line 379: This can maybe be simplified when combining it with `get_kinematic_representation()`

expertsystem/state/particle.py
```diff
@@ -258,6 +264,198 @@ def convert_to_dict(
...
+        if final_state is not None:
+            self.__final_state = self.__import(final_state)
+
+    @property
+    def initial_state(self) -> Optional[List[List[Any]]]:
+        return self.__initial_state
+
+    @property
+    def final_state(self) -> Optional[List[List[Any]]]:
+        return self.__final_state
```
spflueger, line 286: is that really needed? To me only the equality and subset operators are important

expertsystem/state/particle.py
redeboer: This constructor is currently used here:
https://github.com/ComPWA/expertsystem/blob/8b0660eda9d2acc817682dab3c1c5a648d48c455/expertsystem/state/particle.py#L678-L681

expertsystem/state/particle.py
redeboer: Yes, because:
https://github.com/ComPWA/expertsystem/blob/8b0660eda9d2acc817682dab3c1c5a648d48c455/expertsystem/state/particle.py#L670-L671

expertsystem/state/particle.py
redeboer: It is needed in the `__eq__` and subset (`__contains__`) operators when comparing with `other`
https://github.com/ComPWA/expertsystem/blob/9faf80e51c0835b34293c0c782d83fa377f4443a/expertsystem/state/particle.py#L334-L336

expertsystem/state/particle.py
redeboer: 9faf80e5

expertsystem/state/particle.py
redeboer: 9faf80e5

expertsystem/state/particle.py
redeboer: In other words, `__init__` is required to convert the `allowed_final_state_groupings` that the user provides into a `KinematicRepresentation` while `get_kinematic_representation` is required to compare the `StateTransitionGraph` to those `allowed_final_state_groupings`.

expertsystem/state/particle.py
redeboer: https://github.com/ComPWA/expertsystem/pull/245#discussion_r491052122

expertsystem/state/particle.py
redeboer: What could be done is to turn it into a method `from_graph`, like this method:
https://github.com/ComPWA/expertsystem/blob/03725998271da654aace08f109b07d12ee05dcbf/expertsystem/topology.py#L349-L352

expertsystem/state/particle.py
spflueger: ah yeah ok

expertsystem/state/particle.py
spflueger: But a `self.__initial_state` would suffice then

redeboer: Yes, but `other.initial_state` and `other.final_state` won't work

expertsystem/state/particle.py
spflueger: All right. Then lets leave it as it is for now. For the future it would be good to get this merged somehow

expertsystem/state/particle.py
spflueger: interesting, somehow i was thinking in the friend scope of c++. I guess `_initial_state` would be the python equivalent. Well i leave it up to you

redeboer: Yes `_initial_state` would be possible. But then it's still accessible (you will only be warned by the linters). So I guess I'll leave it like this, now both members only have read access.

spflueger (approved):

PR #246, opened 2020-09-17
chore: implement some smaller fixes

redeboer:
See commits for explanations

### DIFF
--- a/tests/unit/topology/test_graph.py
+++ b/tests/unit/topology/test_graph.py
@@ -42,7 +42,7 @@ def visualize_graphs():
     """Render graphs when running this file directly."""
     ui.load_default_particles()
     topology = create_dummy_topology()
-    graphs = test_initialize_graph(topology, None)
+    graphs = test_initialize_graph(topology, io.load_pdg())
     try:
         # pylint: disable=import-error,import-outside-toplevel
         import graphviz  # type: ignore
--- a/tests/unit/topology/test_graph.py
+++ b/tests/unit/topology/test_graph.py
@@ -5,7 +5,40 @@ import pytest
 from expertsystem import io
 from expertsystem import ui
 from expertsystem.state.particle import initialize_graph
-from expertsystem.topology import StateTransitionGraph
+from expertsystem.topology import (
+    InteractionNode,
+    StateTransitionGraph,
+)
+
+
+class TestInteractionNode:
+    @staticmethod
+    def test_constructor_exceptions():
+        dummy_type_name = "type_name"
+        with pytest.raises(TypeError):
+            assert InteractionNode(
+                dummy_type_name,
+                number_of_ingoing_edges="has to be int",  # type: ignore
+                number_of_outgoing_edges=2,
+            )
+        with pytest.raises(TypeError):
+            assert InteractionNode(
+                dummy_type_name,
+                number_of_outgoing_edges="has to be int",  # type: ignore
+                number_of_ingoing_edges=2,
+            )
+        with pytest.raises(ValueError):
+            assert InteractionNode(
+                dummy_type_name,
+                number_of_outgoing_edges=0,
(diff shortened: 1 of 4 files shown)


tests/unit/topology/test_graph.py
```diff
@@ -42,7 +42,7 @@ def visualize_graphs():
     """Render graphs when running this file directly."""
     ui.load_default_particles()
     topology = create_dummy_topology()
-    graphs = test_initialize_graph(topology, None)
+    graphs = test_initialize_graph(topology, io.load_pdg())
```
spflueger: Isn't particle_database a fixture? So then that second argument would not be needed

tests/unit/topology/test_graph.py
```diff
@@ -43,15 +43,7 @@ def visualize_graphs():
...
-    try:
-        # pylint: disable=import-error,import-outside-toplevel
-        import graphviz  # type: ignore
-
-        dot_source = io.dot.convert_to_dot(graphs)
-        vis = graphviz.Source(dot_source)
-        vis.view()
-    except ModuleNotFoundError:
-        pass
+    io.write(graphs, "jpsi_to_gamma_pi0_pi0.gv")
```
spflueger: Can any graphviz dependencies be removed now?

tests/unit/topology/test_graph.py
redeboer: Yeah but a fixture is only loaded when running through `pytest`. This is only there when you directly run the `__main__`, it's a trick to visualise the topology if you do have `graphviz`.

tests/unit/topology/test_graph.py
redeboer: You mean this one?
https://github.com/ComPWA/expertsystem/blob/2bcbb612c78bb633325ae9e7fd97a1cecbecc1a5/setup.cfg#L65

No that one is needed for the documentation.

tests/unit/topology/test_graph.py
spflueger: Ok, but is that not using the fixture then in case of running with pytest?

redeboer: The function `visualize_graphs` is not called through `pytest` at all, because it doesn't start or end in `test`.

tests/unit/topology/test_graph.py
spflueger: Just a general question, in case something can be removed

tests/unit/topology/test_graph.py
redeboer: Note also that this is the reason for the slightly weird construct here
https://github.com/ComPWA/expertsystem/blob/f63f8a6e958cfa74678f1e10eef4d9922c84ffae/tests/unit/topology/test_graph.py#L44-L58

tests/unit/topology/test_graph.py
redeboer: Btw, this `try`-`except` construct was there just to facilitate the conversion of the DOT language into a PDF file, but it was actually a bit annoying to have that PDF pop up each time, so instead I now leave that conversion up to the developer.

Note that we facilitate rendering of a DOT file since #235 through a VSCode extension:
https://github.com/ComPWA/expertsystem/blob/63fa906de5a92381ba38c2d5055d910eaa324961/.vscode/extensions.json#L10

tests/unit/topology/test_graph.py
spflueger: Ah I completely missed that. Ok, so then depending on the purpose of this code I would either promote it to a test or change it into some example code/jupyter notebook (can be handled in a separate PR). However the purpose of the code is not quite clear to me

redeboer: It's mainly just a way to (if you have the right setup) quickly check what the dummy topology looks like

spflueger (approved):

PR #247, opened 2020-09-17
ci: implement pyright and VSCode Pylance

redeboer:
Pylance seems to be the direction in which VSCode is moving
https://devblogs.microsoft.com/python/announcing-pylance-fast-feature-rich-language-support-for-python-in-visual-studio-code/

The original motivation to switch to Pylance was **improved language navigation**, but it also provides diagnostics. In the long run, we may want to implement `pyright` in the CI as well, just like the `npm` packages `cspell` and `markdownlint`.

_Rebased on #246_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,10 +93,12 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install .[dev]
-          sudo npm install -g cspell markdownlint-cli
+          sudo npm install -g cspell markdownlint-cli pyright
       - name: Perform style checks
         run: tox -e sty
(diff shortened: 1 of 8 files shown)


redeboer:
Would be cleaner if `pyright` is actually enforced through CI. However, same as `cspell` and `markdownlint`, `pyright` is an `npm` package and it's hard to integrate this locally into `tox` as well. So I'm trying out a bit how to run these packages from `tox` while not resulting in error messages if you don't have these packages installed, see e.g.:
https://tox.readthedocs.io/en/latest/example/basic.html#ignoring-exit-code

spflueger (approved):

PR #248, opened 2020-09-17
refactor: extract Topology class

redeboer:
Improvements to the `topology` module in order to distinguish the different initialize stages of `StateTransitionGraph`s:

* Extracted `Topology` base class (no node/edge properties) from `StateTransitionGraph`
* Nodes have become a `set`
* Protected nodes and edges of `Topology`
* Type-templated edge properties of `StateTransitionGraph`
* `Edge` has become a `dataclass`

_Rebased on #247_

### DIFF
--- a/expertsystem/state/particle.py
+++ b/expertsystem/state/particle.py
@@ -29,7 +29,7 @@ from expertsystem.data import (
     ParticleCollection,
     Spin,
 )
-from expertsystem.topology import StateTransitionGraph
+from expertsystem.topology import StateTransitionGraph, Topology
 
 
 StateWithSpins = Tuple[str, Sequence[float]]
@@ -460,7 +460,7 @@ class CompareGraphElementPropertiesFunctor:
 
 
 def initialize_graph(  # pylint: disable=too-many-locals
-    empty_topology: StateTransitionGraph,
+    topology: Topology,
     particles: ParticleCollection,
     initial_state: Sequence[StateDefinition],
     final_state: Sequence[StateDefinition],
@@ -475,8 +475,8 @@ def initialize_graph(  # pylint: disable=too-many-locals
                 f"(len({state_definitions}) != len({edge_ids})"
             )
 
-    is_edges = empty_topology.get_initial_state_edges()
-    fs_edges = empty_topology.get_final_state_edges()
+    is_edges = topology.get_initial_state_edges()
+    fs_edges = topology.get_final_state_edges()
     assert_number_of_states(initial_state, is_edges)
     assert_number_of_states(final_state, fs_edges)
 
@@ -488,16 +488,14 @@ def initialize_graph(  # pylint: disable=too-many-locals
     )
 
     attached_is_edges = [
-        empty_topology.get_originating_initial_state_edges(i)
-        for i in empty_topology.nodes
+        topology.get_originating_initial_state_edges(i) for i in topology.nodes
     ]
     is_edge_particle_pairs = __calculate_combinatorics(
         is_edges, initial_state_with_projections, attached_is_edges
     )
 
     attached_fs_edges = [
-        empty_topology.get_originating_final_state_edges(i)
-        for i in empty_topology.nodes
+        topology.get_originating_final_state_edges(i) for i in topology.nodes
     ]
     fs_edge_particle_pairs = __calculate_combinatorics(
         fs_edges,
@@ -512,7 +510,7 @@ def initialize_graph(  # pylint: disable=too-many-locals
             merged_dicts = initial_state_pair.copy()
             merged_dicts.update(fs_pair)
             new_graphs.extend(
-                __initialize_edges(empty_topology, merged_dicts, particles)
+                __initialize_edges(topology, merged_dicts, particles)
             )
 
     return new_graphs
@@ -621,10 +619,11 @@ def __initialize_external_edge_lists(
 
 
 def __initialize_edges(
-    graph: StateTransitionGraph,
+    topology: Topology,
     edge_particle_dict: Dict[int, StateWithSpins],
     particle_db: ParticleCollection,
 ) -> List[StateTransitionGraph]:
+    graph = StateTransitionGraph.from_topology(topology)
     for edge_id, state_particle in edge_particle_dict.items():
         particle_name = state_particle[0]
         particle = particle_db[particle_name]
--- a/expertsystem/state/particle.py
+++ b/expertsystem/state/particle.py
@@ -481,13 +481,12 @@ def initialize_graph(  # pylint: disable=too-many-locals
             f"({len(fs_edges)} !=  {len(final_state)})"
         )
 
-    # check if all initial and final state particles have spin projections set
-    initial_state_with_projections = [
-        __safe_set_spin_projections(x, particles) for x in initial_state
-    ]
-    final_state_with_projections = [
-        __safe_set_spin_projections(x, particles) for x in final_state
-    ]
+    initial_state_with_projections = __safe_set_spin_projections(
+        initial_state, particles
+    )
+    final_state_with_projections = __safe_set_spin_projections(
+        final_state, particles
+    )
 
     attached_is_edges = [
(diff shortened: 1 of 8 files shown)


expertsystem/state/particle.py
```diff
@@ -518,7 +518,7 @@ def assert_number_of_states(
     return new_graphs
 
 
-def __safe_set_spin_projections(
+def _safe_set_spin_projections(
```
spflueger: why is that needed?

tests/unit/topology/test_graph.py
```diff
@@ -58,9 +64,36 @@ def dummy_topology():
...
+        (["J/psi(1S)"], ["gamma", "pi0", "pi0"]),
+        (["J/psi(1S)"], ["K+", "K-", "pi+", "pi-"]),
+        (["e+", "e-"], ["gamma", "pi-", "pi+"]),
+        (["e+", "e-"], ["K+", "K-", "pi+", "pi-"]),
+    ],
+)
+def test_permutations(
+    initial_state, final_state, dummy_topology, particle_database
 ):
+    initial_state_with_spins = _safe_set_spin_projections(
```
spflueger, line 89: Ok, that's downside in the commit based reviews. This answers my question from before. However I'm questioning the `particle` API now. So concretely can the API be changed so that new function can provide this functionality (combined) and calls to  `_safe_set_spin_projection` can be avoided?

expertsystem/state/particle.py
```diff
@@ -542,6 +545,23 @@ def safe_set_spin_projections(
     ]
 
 
+def generate_outer_edge_permutations(
+    topology: StateTransitionGraph,
+    initial_state: Sequence[StateWithSpins],
+    final_state: Sequence[StateWithSpins],
+) -> Generator[Dict[int, StateWithSpins], None, None]:
```
spflueger, line 550: Does this function does not need the `StateWithSpins` right? In other words a particle name or pid would suffice. This connects to the question below

expertsystem/topology.py
```diff
@@ -41,22 +41,23 @@ def __eq__(self, other: object) -> bool:
...
+    ) -> None:
         self.__nodes: Set[int] = set()
         self.edges: Dict[int, Edge] = {}
-        self.node_props: Dict[int, dict] = {}
-        self.edge_props: Dict[int, dict] = {}
-        self.graph_element_properties_comparator: Optional[Callable] = None
+        if nodes is not None:
+            self.__nodes = set(nodes)
+        if edges is not None:
+            self.edges = edges
```
spflueger: This unsafe as the edge could point to non existing nodes etc

expertsystem/topology.py
```diff
@@ -41,22 +41,23 @@ def __eq__(self, other: object) -> bool:
         raise NotImplementedError
 
 
-class StateTransitionGraph:
-    """Graph class that contains edges and nodes.
+class Topology:
+    """Directed graph for a state transition.
```
spflueger, line 37: The thing is that this is not a normal graph of graph theory (CS). I think some comment explaining the difference would be good (maybe even a reference to feynman graphs)

expertsystem/topology.py
```diff
@@ -22,23 +23,12 @@
...
-
-    def __eq__(self, other: object) -> bool:
-        if isinstance(other, Edge):
-            return (
-                self.ending_node_id == other.ending_node_id
-                and self.originating_node_id == other.originating_node_id
-            )
-        raise NotImplementedError
+    ending_node_id: Optional[int] = None
+    originating_node_id: Optional[int] = None
```
spflueger, line 33: Ideal would be to avoid having the state (None, None), which is def invalid

expertsystem/topology.py
```diff
@@ -260,7 +262,10 @@ def swap_edges(self, edge_id1: int, edge_id2: int) -> None:
         self.__edges[edge_id1] = popped_edge_id2
 
 
-class StateTransitionGraph(Topology):
+EdgeType = TypeVar("EdgeType")
```
spflueger, line 265: can be hidden from api

spflueger (commented):
Nice job altogether. I have a few question/comments

expertsystem/state/particle.py
redeboer: It is called in `test_graph.py`

expertsystem/state/particle.py
redeboer: Oh oops, this commit actually belongs to #245

redeboer: I moved the commit to #245. But still, this `generate_outer_edge_permutations` uses the output of `_safe_set_spin_projections`, which is why the properties are `StateWithSpins` at this stage. Can also make it more abstract by using `Any`, because internally this function does not need `StateWithSpins`.

tests/unit/topology/test_graph.py
redeboer: By new function, do you mean that `_safe_set_spin_projections` now takes a list? I'm not sure how `_safe_set_spin_projection can be avoided, because we want to allow the user to not specify spin projections in case all allowed spin projections should be considered.

expertsystem/topology.py
redeboer: Currently not possible because of the way in which `Topology` is built up elsewhere in the system through `add_edge`:

https://github.com/ComPWA/expertsystem/blob/5bca7a79be63fe5f18b2c10e151dce66b3f6dd62/expertsystem/topology.py#L91

expertsystem/topology.py
redeboer: 5bca7a79be63fe5f18b2c10e151dce66b3f6dd62

expertsystem/topology.py
redeboer: ae844f3e

tests/unit/topology/test_graph.py
spflueger: Well the calls made here are really similar to what the `initialize_graphs` function is doing internally. So I'm saying, maybe instead `initialize_graphs` could be called (or some other new function, which would be a subpart of initialize graphs)

expertsystem/state/particle.py
spflueger: making it more abstract sounds good to me

tests/unit/topology/test_graph.py
redeboer: > So I'm saying, maybe instead initialize_graphs could be called (or some other new function, which would be a subpart of initialize graphs)

I don't understand. `_safe_set_spin_projections` is called from `initialize_graphs` itself.

expertsystem/topology.py
spflueger: Ok, lets put that on the Todolist. Lets open a new Epic issue for that

redeboer: Should that really be an epic?

spflueger: those two classes are still quite unsafe

redeboer: Yes that's true. I'm currently working on the `Topology.verify` method, to address at least some of those issues.

spflueger: We can also make the way of building an invalid graph impossible, by changing the interface on how to add edges and nodes etc. For example these rules are quite solid:
- an Edge has to be attached to at least one node
- a new Node has to be attached to at least one edge (except the first node/seed node)

tests/unit/topology/test_graph.py
spflueger: Hmm, why i can't see this change anymore? The old code seems to use it the way I was suggesting

expertsystem/topology.py
redeboer: #250

tests/unit/topology/test_graph.py
redeboer: 7d780ec3

expertsystem/topology.py
redeboer: Discussion moved to #250

spflueger (approved):

PR #249, opened 2020-09-17
ci!: add cspell and markdownlint to tox

redeboer:
This PR was triggered by https://github.com/ComPWA/expertsystem/pull/247#issuecomment-694197252

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,6 +9,8 @@ envlist =
     sty,
 
 [testenv]
+description =
+    Run all fast unit tests
 whitelist_externals =
     pytest
 commands =
@@ -18,8 +20,9 @@ commands =
         --cov-report=xml \
         --cov=expertsystem
 
-; Allows to quickly build html with `tox -e doc` from the root dir
 [testenv:doc]
+description =
+    Build documentation and API through Sphinx
 changedir = doc
 whitelist_externals =
     make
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -46,11 +46,14 @@ commands =
...
     SKIP = mypy
 whitelist_externals =
+    bash
     mypy
     pre-commit
 commands =
     mypy .  # run separately because of potential caching problems
     pre-commit run {posargs} -a
+    - bash -ec "cspell $(git ls-files)"
+    - bash -ec "markdownlint ."
```
spflueger, line 56: How do we guarantee that cpsell and markdownlint are installed?

redeboer: Non-zero exit codes of commands starting with `-` are ignored, see
https://tox.readthedocs.io/en/latest/example/basic.html#ignoring-exit-code

Note, however, that this also means `tox -e sty` does not fail if `cspell` or `markdownlint` do emit errors, so you would have to see this in the terminal. This is also the reason why both `cspell` and `markdownlint` are still called separately in GitHub Actions.

I've also not added `pyright`, because it doesn't find the imports when executed through `tox`.

spflueger: i see. thx for the clarification

spflueger (approved):

PR #250, opened 2020-09-17
feat: implement Topology.verify and clean up

redeboer:
Triggered by https://github.com/ComPWA/expertsystem/pull/248#discussion_r490356949 and https://github.com/ComPWA/expertsystem/pull/248#discussion_r490348892. Doesn't really cleanly it, but the `Topology` class becomes safer overall.

_Rebased on #248_


expertsystem/topology.py
```diff
@@ -152,11 +154,40 @@ def get_originating_node_list(
             self.edges[edge_id].originating_node_id for edge_id in edge_ids
         ]
 
-    def verify(self) -> bool:  # pylint: disable=no-self-use
-        """Verify if the graph is connected.
+    def verify(self) -> None:
```
spflueger, line 157: can be made private

expertsystem/topology.py
```diff
@@ -152,11 +154,40 @@ def get_originating_node_list(
...
+                    f"Edge nr. {edge_id} is not connected to any node ({edge})"
+                )
+            if not {edge.originating_node_id, edge.ending_node_id} <= node_ids:
+                raise ValueError(
+                    f"{edge} (ID: {edge_id}) has non-existing node IDs.\n"
+                    f"Available node IDs: {self.nodes}"
+                )
+        self.check_isolated_nodes()
+
+    def check_isolated_nodes(self) -> None:
```
spflueger, line 175: private. actually you can do the same for all of the other functions that are not required to be public

expertsystem/topology.py
```diff
@@ -32,6 +32,11 @@ class Edge:
     ending_node_id: Optional[int] = None
     originating_node_id: Optional[int] = None
 
+    def get_connected_nodes(self) -> Set[int]:
+        connected_edges = {self.ending_node_id, self.originating_node_id}
```
spflueger, line 36: *connected_nodes

expertsystem/topology.py
```diff
@@ -152,11 +154,40 @@ def get_originating_node_list(
...
-    def verify(self) -> bool:  # pylint: disable=no-self-use
-        """Verify if the graph is connected.
+    def verify(self) -> None:
+        """Verify if there are no dangling edges or nodes."""
+        node_ids = self.nodes | {None}  # trick to ignore None
+        for edge_id, edge in self.edges.items():
+            if (
+                edge.originating_node_id is None
+                and edge.ending_node_id is None
+            ):
```
spflueger, line 164: You could call `get_connected_nodes()` here

expertsystem/topology.py
```diff
@@ -152,11 +154,40 @@ def get_originating_node_list(
...
+                raise ValueError(
+                    f"Edge nr. {edge_id} is not connected to any node ({edge})"
+                )
+            if not {edge.originating_node_id, edge.ending_node_id} <= node_ids:
```
spflueger, line 168: same as above

expertsystem/topology.py
```diff
@@ -152,11 +154,40 @@ def get_originating_node_list(
...
+                    f"Available node IDs: {self.nodes}"
+                )
+        self.check_isolated_nodes()
+
+    def check_isolated_nodes(self) -> None:
+        if len(self.nodes) < 2:
+            return
+        for node_id in self.nodes:
+            surrounding_nodes = self.get_surrounding_nodes(node_id)
+            surrounding_nodes.discard(node_id)
```
spflueger, line 180: would be cleaner to discard that in the `get_surrounding_nodes` function

expertsystem/topology.py
```diff
@@ -257,6 +290,14 @@ def get_originating_initial_state_edges(self, node_id: int) -> List[int]:
             temp_edge_list = new_temp_edge_list
         return edge_list
 
+    def get_surrounding_nodes(self, node_id: int) -> Set[int]:
```
spflueger, line 293: private

tests/unit/state/test_graph_combinatorics.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@pytest.mark.parametrize(
+    "final_state_groupings", [[[["pi0", "pi0"]]], [[["gamma", "pi0"]]],],
+)
+def test_initialize_graph(
+    final_state_groupings, dummy_topology, particle_database
+):
+    graphs = initialize_graph(
+        dummy_topology,
```
spflueger, line 13: Idea: the dummy topology is not even needed as a fixture. You could create a 3 body final state decay topology with the builder (there is only one topology).

```python
two_body_decay_node = InteractionNode("TwoBodyDecay", 1, 2)
simple_builder = SimpleStateTransitionTopologyBuilder(
    [two_body_decay_node]
￼)
all_graphs = simple_builder.build_graphs(1, 3)
```

tests/unit/test_topology.py
```diff
@@ -0,0 +1,112 @@
...
+        assert topology.is_valid()
+        topology.add_node(2)
+        assert not topology.is_valid()
+        topology.add_edges([5])
+        assert not topology.is_valid()
+        topology.attach_edges_to_node_ingoing([5], 2)
+        assert not topology.is_valid()
+
+    @staticmethod
+    def test_check_isolated_nodes(dummy_topology):
```
spflueger, line 48: It would propose to remove `is_valid` and `verify` will be the only public interface function (regarding verification). Then various toplogies can be created and verified. An depending on the specifc Errors raised a more granular way of testing is possible.
That means this test can be "merged" with the one above

tests/unit/test_topology.py
```diff
@@ -0,0 +1,112 @@
...
+            assert Topology(edges={0: Edge(None, 1)})
+        with pytest.raises(ValueError):
+            assert Topology(nodes={0}, edges={0: Edge(1, None)})
+        with pytest.raises(ValueError):
+            assert Topology(edges={0: Edge(None, None)})
+        topology = Topology(nodes={0}, edges={0: Edge(0, None)})
+        assert len(topology.nodes) == 1
+
+    @staticmethod
+    def test_is_valid(dummy_topology):
```
spflueger, line 37: I would rather make this into a test that creates various topologies as inputs as this dummy topology, for example using the TopologyBuilder. Because its very hard to understand this test if you dont know what the dummy topology is.

spflueger (changes requested):

tests/unit/state/test_graph_combinatorics.py
redeboer: Mwa, I want to keep the tests for `Topology` isolated form the test for the topology builders.

tests/unit/test_topology.py
redeboer: 41cc7c1b

tests/unit/test_topology.py
redeboer: 81bc9f5f

expertsystem/topology.py
redeboer: 068686a5

expertsystem/topology.py
redeboer: e4e05576

expertsystem/topology.py
redeboer: 3541b385 and 8cee9310

expertsystem/topology.py
redeboer: 3541b385 and 8cee9310

expertsystem/topology.py
redeboer: 31f9aaad

expertsystem/topology.py
redeboer: 810f0406

expertsystem/topology.py
redeboer: c4197296 and 81bc9f5f

tests/unit/test_topology.py
redeboer: But it might be better to make the `dummy_topology` a `three_body_topology` and tests a few more of the `Topology` methods

spflueger: Yeah that's def a better name for it

redeboer: 38464a68 and bf1b1e66

redeboer:
Completed the unit tests for `Topology`. Also tried to return `Set` in the getter methods, but that messes up the combinatorics.

tests/unit/state/test_graph_combinatorics.py
```diff
@@ -0,0 +1,36 @@
...
+from expertsystem.state.particle import initialize_graph
+from expertsystem.topology import (
+    InteractionNode,
+    SimpleStateTransitionTopologyBuilder,
+    Topology,
+)
+
+
+@pytest.fixture(scope="package")
+def two_body_decay() -> Topology:
```
spflueger, line 14: This is actually a 1 to 3 body decay topology. So the name is a bit misleading to me

tests/unit/test_topology.py
```diff
@@ -0,0 +1,212 @@
...
+            0: Edge(0, None),
+            1: Edge(1, 0),
+            2: Edge(None, 0),
+            3: Edge(2, 1),
+            4: Edge(None, 1),
+            5: Edge(None, 2),
+            6: Edge(None, 2),
+        },
+    )
+    io.dot.write(topology, "four_body_decay.gv")
```
spflueger, line 40: why the write?

spflueger (commented):
Almost there, two more questions

tests/unit/test_topology.py
redeboer: Just a helper in case you want to visualise

tests/unit/state/test_graph_combinatorics.py
redeboer: e9715e8
Now I'm thinking whether it wouldn't be better to test a 2-to-3 body decay instead of a 1-to-4 in `test_topology.py::TestTopology` :thinking:

redeboer: 5333e4fc

spflueger: ok good

tests/unit/test_topology.py
spflueger: Hmm i would separate that though (like one fixture depending on another), is it actually needed right now?

redeboer: Nah it was just a trick for me to check whether it looked like what I had in mind :grimacing:

spflueger: Then i would just remove it

redeboer: 8b83b09a

spflueger (approved):
Good job!

PR #252, opened 2020-09-21
style: upgrade and apply black and prettier

redeboer:
Black 20.8:
- Support for Python 3.8
- Bug fix "type: ignore" comment
- Lists ending in "," get a forced line break


spflueger (approved):
Black also formats the html file?

redeboer:
> Black also formats the html file?

No, that's Prettier

PR #260, opened 2020-09-23
feat: implement ParticleCollection.filter

redeboer:
See [commits](https://github.com/ComPWA/expertsystem/pull/260/commits) for additional changes


expertsystem/data.py
```diff
@@ -207,6 +207,23 @@ def __post_init__(self) -> None:
...
+            value = getattr(self, field.name)
+            if value is None:
+                continue
+            if field.name not in ["mass", "spin", "isospin"] and value == 0:
+                continue
+            if isinstance(value, str):
+                value = f'"{value}"'
+            output_string += f"\n    {field.name}={value},"
+        output_string += "\n)"
+        return output_string
```
spflueger, line 227: Why is this needed? If we want some better "string" representation of this for prettier printing, then `__str__` should be overwritten. Btw, I would assume there is a reason why dataclasses list default values on purpose, like guaranteeing same instance even if default values change.

expertsystem/data.py
```diff
@@ -288,13 +289,13 @@ def compute_isospin_projection(  # pylint: disable=too-many-arguments
 class ParticleCollection(abc.Mapping):
     """Safe, `dict`-like collection of `.Particle` instances."""
 
-    def __init__(
-        self, particles: Optional[Dict[str, Particle]] = None
-    ) -> None:
+    def __init__(self, particles: Optional[Iterable[Particle]] = None) -> None:
```
spflueger, line 293: I'm thinking again about the data structure of the `ParticleCollection` itself. Why can't it be a set again?

expertsystem/data.py
```diff
@@ -352,29 +353,28 @@ def find(self, search_term: Union[int, str]) -> Particle:
             f"Cannot search for a search term of type {type(search_term)}"
         )
 
-    def find_subset(
-        self, search_term: Union[int, str]
+    def filter(  # noqa: A003
+        self, function: Callable[[Particle], bool]
```
spflueger, line 357: So its `ParticleCollection.filter(lambda p: p.charge == 1)` vs `[p for p in ParticleCollection if p.charge == 1]`

In my view the filter function is unnecessary, if `ParticleCollection` can be a basic datatype like a `set` (see previous comment).

spflueger (commented):
Looks nice, I have a fundamental question regarding the `ParticleCollection` type though.

expertsystem/data.py
redeboer: It's mainly for the faster lookup. We assume most look up by name, so that's why we use a `dict` with names as keys. Otherwise you'd need to loop over the set somehow.

expertsystem/data.py
redeboer: It's actually mainly for the Jupyter notebooks. The `Particle` class is an important interface to the outside, so you want it to render nicely when doing for instance:
```python
from expertsystem import io
pdg = io.load_pdg()
pdg["gamma"]
```
in a notebook. So for instance, not have all those `isospin=None` statements.

Note: in this example, Jupyter notebook calls `Particle.__repr__`, not `Particle.__str__`.

redeboer: Also note that he `__repr__` method is checked with `eval` for all instances in the PDG.

expertsystem/data.py
redeboer: https://github.com/ComPWA/expertsystem/pull/260#discussion_r493306662

redeboer: Also see
https://pwa--260.org.readthedocs.build/projects/expertsystem/en/260/usage/particles.html
(unfortunately no cell output in the PR rendering on RTD, while that's actually the point of the modifications to `ParticleCollection.__repr__` and `Particle.__repr__`.

expertsystem/data.py
spflueger: i think we need `_repr_pretty_` then, see https://stackoverflow.com/questions/41453624/tell-ipython-to-use-an-objects-str-instead-of-repr-for-output

expertsystem/data.py
spflueger: True, thanks

expertsystem/data.py
spflueger: So since `ParticleCollection` is not a primitive type (like `set`), this is totally fine. Because we want to have a ParticleCollection again after the filtering.

expertsystem/data.py
redeboer: Ok I tried it out a bit, but can't get it to work nicely (especially when printing `ParticleCollection`).

redeboer: Apart from that, this PR was originally motivated by my attempts to create some widgets through Sphinx. I'm currently experimenting with [`sphinx-thebe`](https://sphinx-thebe.readthedocs.io/en/latest/use.html), which uses Python instead of Python (that is, you need a `print` statement to render output).

spflueger: Don't know, just seems fishy to me, as in it is a misuse of `__repr__`. If it would be the purpose to serialize objects via `__repr__` and default arguments are not shown, then when the default arguments change backwards compatibility is lost. Not sure if this is the purpose of `__repr__` though

redeboer: It may indeed be fishy, depends on if we understand Python's intention correctly. Still, an `eval` test is performed for all instances in the particle database from the PDG, so the implementation is quite safe (also with regard to default arguments).

spflueger (approved):
I leave the `__repr__` issue up to you, to figure out if this an ok thing to do.

Issue #173, opened 2020-08-06
Use Particle and spin projection within STM._determine_node_settings


spflueger:
Similar to the other Issues in the Epic, use the `Particle` class more in the implementation and use the adapter to convert it to a nested dict structure to keep the interface stable.

spflueger:
Also closing this in favor of #244 which promotes the use of `Particle` together with a spin projection

spflueger:
Renaming instead of closing

PR #261, opened 2020-09-23
feat: add Particle.is_lepton method

redeboer:
While the title of this PR was the original motivation (it is implemented in #173), the bulk of this PR is actually the reorganisation of the tests for the `data` module. This allowed to spot some mistakes and clean up duplicate tests. The convention followed there is explained in a new section in the contribute page.

Some other improvements:
* The test of `Particle.__repr__` didn't do anything useful, but that has been fixed. This is relevant for the discussion in #260.
* Test collection warnings are treated now treated as errors, because there's a risk certain tests you implemented are not run.
* For VSCode: the `__init__.py` files won't be shown in search or when using code navigation (e.g. `Ctrl+P`).
* Test coverage for the `data` module has improved slightly.


tests/unit/test_data.py
```diff
@@ -75,8 +75,8 @@ def test_exceptions(magnitude, projection):
 
 class TestParticle:
     @staticmethod
-    def test_repr(particle_database):
-        for particle in particle_database:
+    def test_repr(particle_database: ParticleCollection):
+        for particle in particle_database.values():
```
spflueger, line 100: Just a comment, the `values()` seems kind of redundant. I guess the `ParticleCollection` interface is `add`, `find` and iteration. Something for another PR though

expertsystem/data.py
```diff
@@ -296,10 +296,14 @@ class ParticleCollection(abc.Mapping):
...
-            if isinstance(particles, (list, set, tuple)):
-                self.__particles.update(
-                    {particle.name: particle for particle in particles}
+            if not isinstance(particles, (list, set, tuple)):
+                raise ValueError(
+                    f"Cannot construct a {self.__class__.__name__} "
+                    f"from a {particles.__class__.__name__}"
                 )
+            self.__particles.update(
+                {particle.name: particle for particle in particles}
```
spflueger, line 305: Improvement suggestion: Actually there is no runtime check on what `particle` is. So theoretically any kind of object can be injected here as long as it has a attribute `name`.

redeboer: 26fae1f1
I don't want to slow down construction too much though, that's why I put it in the comprehension

tests/unit/test_data.py
redeboer: The reason is that the `ParticleCollection` mimics an (immutable) `Mapping`:
https://github.com/ComPWA/expertsystem/blob/26fae1f1b407ad870ec227635e226fa19d1e9a77/expertsystem/data.py#L293

spflueger (approved):

PR #262, opened 2020-09-23
refactor: remove dicts in determine_node_settings

### LINKED ISSUE #173: Use Particle and spin projection within STM._determine_node_settings

redeboer:
Closes #173

### DIFF
--- a/expertsystem/topology.py
+++ b/expertsystem/topology.py
@@ -320,9 +320,7 @@ class StateTransitionGraph(Topology, Generic[_EdgeType]):
 
     def __repr__(self) -> str:
         return (
-            f"{self.__class__.__name__}()"
-            f"\n    nodes: {self.nodes}"
-            f"\n    edges: {self.edges}"
+            f"{self.__class__.__name__}(nodes={self.nodes}, edges={self.edges})"
             f"\n    node props: {self.node_props}"
             f"\n    node props: {self.edge_props}"
         )
--- a/expertsystem/topology.py
+++ b/expertsystem/topology.py
@@ -322,7 +322,7 @@ class StateTransitionGraph(Topology, Generic[_EdgeType]):
         return (
             f"{self.__class__.__name__}(nodes={self.nodes}, edges={self.edges})"
             f"\n    node props: {self.node_props}"
-            f"\n    node props: {self.edge_props}"
+            f"\n    edge props: {self.edge_props}"
         )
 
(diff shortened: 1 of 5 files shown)


expertsystem/topology.py
```diff
@@ -320,9 +320,7 @@ def __init__(
 
     def __repr__(self) -> str:
         return (
-            f"{self.__class__.__name__}()"
-            f"\n    nodes: {self.nodes}"
-            f"\n    edges: {self.edges}"
+            f"{self.__class__.__name__}(nodes={self.nodes}, edges={self.edges})"
             f"\n    node props: {self.node_props}"
             f"\n    node props: {self.edge_props}"
```
spflueger, line 325: should be `edge props:`

expertsystem/ui/__init__.py
```diff
@@ -237,20 +239,33 @@ def _create_seed_graphs(
...
+                            edge_id
+                        ] = particle_with_spin_projection_to_dict(  # type: ignore
+                            edge_property
+                        )
+
+        if isinstance(instance, list) and len(instance) > 0:
+            if isinstance(instance[0], StateTransitionGraph):
+                convert_edges_in_graph(instance)  # type: ignore
+            elif isinstance(instance[0], tuple):
+                convert_edges_in_graph([g for g, _ in instance])  # type: ignore
```
spflueger, line 265: does this work? since a tuple is immutable it should be not possible to alter it in place

spflueger (changes requested):
Nice. Just two small issues

expertsystem/topology.py
redeboer: 131b22cf
Actually this `__repr__` is still nonsensical and it's also still quite obscure when printing e.g. a `list` of `StateTransitionGraphs` (namely the solutions the STM finds), but a for a better `__repr__` we may have to reconsider the constructor of this class (perhaps even implement a `__new__`).

expertsystem/ui/__init__.py
redeboer: Yes, it's a `tuple` of two, so you unpack the first into `g` and the second into `_`

expertsystem/topology.py
spflueger: Thats true...

Issue #254, opened 2020-09-22
Epic: remove all nested dict structures


redeboer:

redeboer:
Epic closed :tada:

PR #263, opened 2020-09-23
refactor: convert to dict after int. settings

### LINKED ISSUE #174: Use Particle and spin projection in create_interaction_setting_groups
As the last step of this epic also use `Particle` in the implementation of `create_interaction_setting_groups`. This will make the whole function `prepare_graphs` of the STM to work with the new class instead of nested dicts. Use the adapter to return a nested dict structure again to keep the interface stable.

redeboer:
Closes #174

### DIFF
--- a/expertsystem/ui/__init__.py
+++ b/expertsystem/ui/__init__.py
@@ -202,10 +202,10 @@ class StateTransitionManager:  # pylint: disable=too-many-instance-attributes
         seed_graphs = self._create_seed_graphs(topology_graphs)
         graph_node_setting_pairs = self._determine_node_settings(seed_graphs)
         # create groups of settings ordered by "probability"
-        self._convert_edges_to_dict(graph_node_setting_pairs)
         graph_settings_groups = create_interaction_setting_groups(
             graph_node_setting_pairs
         )
+        self._convert_edges_to_dict(graph_node_setting_pairs)
         return graph_settings_groups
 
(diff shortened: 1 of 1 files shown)


expertsystem/ui/__init__.py
```diff
@@ -263,6 +263,10 @@ def convert_edges_in_graph(
                 convert_edges_in_graph(instance)  # type: ignore
             elif isinstance(instance[0], tuple):
                 convert_edges_in_graph([g for g, _ in instance])  # type: ignore
+        elif isinstance(instance, dict):
+            graph_settings_groups = instance
+            for graphs in graph_settings_groups.values():
+                convert_edges_in_graph([g for g, _ in graphs])
```
redeboer, line 269: @spflueger While this addition seems trivial, I have my doubts about the implementation, because I believe this conversion over the `values` of this
```python
Dict[Strength, List[Tuple[StateTransitionGraph[ParticleWithSpin], Dict[int, InteractionNodeSettings]]]]
```
(namely, the `graph_settings_groups`) creates a new `list` of graphs for each strength.

redeboer: Still, even if this is true, this code is temporary and should be moved out asap once #254 has been resolved.

spflueger: yeah the follow up PR to this will already remove that conversion

spflueger (approved):

PR #264, opened 2020-09-23
docs!: add interactive examples page

redeboer:
This creates [a new page](https://pwa--264.org.readthedocs.build/projects/expertsystem/en/264/interactive.html) where we can expose certain features of the `expertsystem` as interactive widges using [`sphinx-thebe`](https://pypi.org/project/sphinx-thebe). Currently, there is only a `ParticleCollection` example, but it can serve as a small online search engine for PDG values.


.gitignore
```diff
@@ -41,3 +41,8 @@ prof/
 # Settings
 .idea/
 **.code-workspace
+
+# Hide 'trick' files from file navigation
+/doc/tox.ini
+/expertsystem/solvers/constraint/tox.ini
+/tests/tox.ini
```
spflueger, line 48: Since these files are under version control, I don't understand why they have to be excluded

redeboer: Ah yes, this is actually from an older attempt. The correct way (at least for VSCode) is 48832b8f

spflueger (approved):

PR #269, opened 2020-09-25
refactor: split state.particle module

redeboer:
The responsibilities of the `expertsystem.state.particle` module have become a bit unclear. It therefore has been split up and renamed into:

- `expertsystem.nested_dicts`, which takes care of the old `dict` structure as long as that it still there. Note that for type hints in Sphinx, this module is not private, but in the documentation it is made clear this module will be phased out.
- `expertsystem.state.combinatorics` containing the functions that generate permutations on the outer edges of the topologies.
- `expertsystem.state.properties` which contains functions to extract and modify edge and node properties.


spflueger (approved):
Looks all right.

PR #270, opened 2020-09-25
feat: implement get_intermediate_states

redeboer:
The `print_intermediate_states` [defined in the quickstart](https://pwa.readthedocs.io/projects/expertsystem/en/latest/usage/quickstart.html#1.3.-Find-solutions) has been moved to the `ui` module (slightly modified form).


examples/quickstart.ipynb
```diff
@@ -56,10 +56,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from expertsystem.ui import (\n",
-    "    InteractionTypes,\n",
-    "    StateTransitionManager,\n",
-    ")"
+    "from expertsystem.ui import InteractionTypes, StateTransitionManager"
```
spflueger, line 59: why does this change, where the brackets removed or did black do this automatically?

expertsystem/ui/__init__.py
```diff
@@ -459,3 +465,26 @@ def _particle_with_spin_projection_to_dict(
...
+def get_intermediate_state_names(
+    solutions: List[StateTransitionGraph],
+) -> Set[str]:
+    """Extract the names of the intermediate states in the solutions."""
+    intermediate_states = set()
+    for graph in solutions:
+        for edge_id in graph.get_intermediate_state_edges():
+            edge_property = graph.edge_props[edge_id]
+            if isinstance(edge_property, dict):
+                intermediate_states.add(edge_property["Name"])
```
spflueger, line 479: we have to remember and clean up this one at the end. I would almost prefer if currently just the dict version is implemented (as thats whats currently returned by the solvers). Then once that is fixed its impossible to miss this change

examples/quickstart.ipynb
redeboer: Black formatting, apparently the line was too long. Not sure why it wasn't formatted before (haven't found a way to format notebooks from the terminal and integrate into the CI).

spflueger: hmm i see

expertsystem/ui/__init__.py
redeboer: That's a good point. I'm working on moving this function to `state.properties`, but now I noticed that I messed up something in #269: the idea was to have that PR split `particles` into `nested_dicts`, `state.properties` and `state.combinatorics`... Follow-up soon

redeboer: Btw, there are more of these 'backward compatible' functions, e.g. the ones introduced in #262.

spflueger: Yes, if both cases are needed, then its of course unavoidable. I don't know the best way around this. My idea is to just spot those old unused cases via code coverage

spflueger (approved):

redeboer:
Want to rebase this one on #272 and move `get_intermediate_state_names` to `state.properties`. Most of the remaining nested `dict` functionality is there (the bulk is of course under `nested_dict`).

redeboer:
https://github.com/ComPWA/expertsystem/pull/270#issuecomment-698810019
Oh nvm, actually this function is too much of a facade to be placed into `state.properties`.

However, I still want #272 to pass first to keep the eventual squashed commit of that one as a follow-up to the one of #269.

PR #271, opened 2020-09-25
feat: improve Parity class

redeboer:
I had some problems when using the `ParticleCollection` in some external notebook when doing some list comprehensions involving parity. This PR makes the `Parity` class behave more as expected.

### DIFF
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -34,15 +34,13 @@ class Parity(abc.Hashable):
         return self.__value == other
 
     def __gt__(self, other: Any) -> bool:
-        if isinstance(other, Parity):
-            return self.value > other.value
         return self.value > int(other)
 
     def __ge__(self, other: Any) -> bool:
         return self > other or self == other
 
     def __lt__(self, other: Any) -> bool:
-        return not self >= other
+        return other > self.value
 
     def __le__(self, other: Any) -> bool:
         return not self > other
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -3,6 +3,7 @@
 import logging
 from collections import abc
 from dataclasses import dataclass, fields
+from functools import total_ordering
 from typing import (
     Any,
     Callable,
@@ -19,6 +20,7 @@ from typing import (
 )
 
 
+@total_ordering
 class Parity(abc.Hashable):
     """Safe, immutable data container for parity."""
 
@@ -35,15 +37,6 @@ class Parity(abc.Hashable):
 
     def __gt__(self, other: Any) -> bool:
         return self.value > int(other)
-
-    def __ge__(self, other: Any) -> bool:
-        return self > other or self == other
-
-    def __lt__(self, other: Any) -> bool:
-        return other > self.value
-
-    def __le__(self, other: Any) -> bool:
-        return not self > other
 
     def __int__(self) -> int:
         return self.value
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -4,6 +4,7 @@ import logging
 from collections import abc
 from dataclasses import dataclass, fields
 from typing import (
+    Any,
     Callable,
     Dict,
     ItemsView,
@@ -31,6 +32,20 @@ class Parity(abc.Hashable):
         if isinstance(other, Parity):
             return self.__value == other.value
         return self.__value == other
+
+    def __gt__(self, other: Any) -> bool:
+        if isinstance(other, Parity):
+            return self.value > other.value
+        return self.value > int(other)
+
+    def __ge__(self, other: Any) -> bool:
+        return self > other or self == other
+
+    def __lt__(self, other: Any) -> bool:
+        return not self >= other
+
+    def __le__(self, other: Any) -> bool:
+        return not self > other
 
     def __int__(self) -> int:
         return self.value
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -18,7 +18,7 @@ from typing import (
 )
 
 
-class Parity:
+class Parity(abc.Hashable):
     """Safe, immutable data container for parity."""
 
     def __init__(self, value: Union[float, int, str]) -> None:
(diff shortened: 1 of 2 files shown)


expertsystem/data.py
```diff
@@ -32,12 +33,29 @@ def __eq__(self, other: object) -> bool:
...
+
     def __int__(self) -> int:
         return self.value
 
     def __neg__(self) -> "Parity":
         return Parity(-self.value)
 
+    def __hash__(self) -> int:
+        return self.__value
+
```
redeboer, line 58: Side note: I had problems with the (previously) missing `__hash__` method when creating a `set` of `tuples` with particle info (in this case a set of tuples of parities with spins).

spflueger: I can't really follow the need for this functionality?

expertsystem/data.py
```diff
@@ -32,12 +33,29 @@ def __eq__(self, other: object) -> bool:
...
+    def __ge__(self, other: Any) -> bool:
+        return self > other or self == other
+
+    def __lt__(self, other: Any) -> bool:
+        return not self >= other
+
+    def __le__(self, other: Any) -> bool:
+        return not self > other
+
```
spflueger, line 49: why is this needed?

expertsystem/data.py
redeboer: I tried to do something like this:
```python
>>> from expertsystem import io
p.parity, p.spin) for p in pdg.values()}
>>> pdg = io.load_pdg()
>>> {(p.parity, p.spin) for p in pdg.values()}
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 1, in <setcomp>
TypeError: unhashable type: 'Parity'
```

expertsystem/data.py
redeboer: Just to have expected behaviour like:
```python
from expertsystem import io
pdg = io.load_pdg()
{p for p in pdg if p.parity < 0}
```
One could wonder about the `<=` etc but I would have that as well if you introduce this '`int`-like behaviour.

expertsystem/data.py
spflueger: Ok, I'm not sure what the use case for doing something like this would be. I guess a user wants to reduce the particle to some important qns. Btw, you could still do `(p.parity.value,...`

expertsystem/data.py
spflueger: I just don't see a use case when someone wants to do such an operation.

expertsystem/data.py
redeboer: That is true. Or, alternatively, `int(p.parity)`.

expertsystem/data.py
redeboer: The `ParticleCollection.filter` function has become a bit of a quantum number search engine (see [interactive exampels](https://pwa.readthedocs.io/projects/expertsystem/en/latest/interactive.html#particle-database)). So I want that function to work as conveniently as possible.
```python
pdg.filter(lambda p: p.parity < 0 and p.spin == 1)
```
But I'm aware that there is a risk: when does convenience become unexpected behaviour?

expertsystem/data.py
redeboer: See also https://github.com/ComPWA/expertsystem/pull/271#discussion_r494829408

expertsystem/data.py
spflueger: Ok thats maybe a bad example as this would be the same as `p.parity == -1`, but I see your point now.
Like if you want unknown and -1 parity you could do <= 0 or < 1 instead of `parity == -1 and parity == 0` (or None). And its a bit more convenient than having to write `parity.value`

redeboer: This point is however related to #259, which would make `Parity` a more complicated object. Right now, it's just a 'read-only' `int` that can only be ±1.

expertsystem/data.py
```diff
@@ -32,12 +33,29 @@ def __eq__(self, other: object) -> bool:
...
+    def __gt__(self, other: Any) -> bool:
+        if isinstance(other, Parity):
+            return self.value > other.value
+        return self.value > int(other)
+
+    def __ge__(self, other: Any) -> bool:
+        return self > other or self == other
+
+    def __lt__(self, other: Any) -> bool:
+        return not self >= other
```
spflueger, line 45: im not sure about standard python procedure here, but I'm quite used to just swapping the operands, so here
`self < other` -> `other > self`

redeboer: You mean `self < other` instead of `not self >= other`? But that would cause an infinite recursion

spflueger: No so the general idea is you define either `<` or `>`, and then you just switch the arguments if you want the opposite.
So if you defined greater than (>) then you would define less than (A < B) as `return B > A`.

redeboer: Not sure what you mean. The methods implemented here were required to get the following tests working:
https://github.com/ComPWA/expertsystem/blob/148b9419d5001f5f18447ea2346699052c2d33e7/tests/unit/test_data.py#L81-L89

spflueger: I'm saying
```python3
def __lt__(self, other: Any) -> bool:
        return other > self
```
looks better than
```python3
def __lt__(self, other: Any) -> bool:
        return not self >= other
```

redeboer: Ah ok. But still, it causes a max recursion:
```
expertsystem/data.py:45: in __lt__
    return other > self
expertsystem/data.py:45: in __lt__
    return other > self
E   RecursionError: maximum recursion depth exceeded in comparison
!!! Recursion detected (same locals & position)
```
(this is done on 148b9419 with the exact replacement as you describe)

spflueger: Does not make sense to me. I would say you made a mistake in your implementation then. So you should define 
```python3
def __gt__(self, other: Any) -> bool:
        return self.value > other.value 
```
(Note i simplified the implementation)
So not keep pointing one function to another. So you define one function and make all the other ones point to this one

redeboer: ca40fb25

spflueger: Ok just for clarification: the reason for the endless loop was the comparison for an `int` with a `Parity`

spflueger (approved):

PR #274, opened 2020-09-28
feat!: provide main facade functions

### LINKED ISSUE #343: Provide facade functions in the top module
In the [renewed API](https://pwa.readthedocs.io/projects/expertsystem/en/0.6.0/api.html), definitions have been moved down to the three main submodules. The idea is that they allow for more flexible usage (e.g. adding final state groupings or modyfing the interaction settings). For regular usage, some 'workflow' facades in the main `expertsystem` module would be nice though.

As an example that users currently have to copy paste for initial usage, is the one below.

```python
from expertsystem.reaction import InteractionTypes, StateTransitionManager

stm = StateTransitionManager(
    initial_state=["psi(2S)"],
    final_state=["gamma", "eta", "eta"],
    formalism_type="helicity",
)
stm.set_allowed_interaction_types([InteractionTypes.EM, InteractionTypes.Strong])
graph_interaction_settings_groups = stm.prepare_graphs()
result = stm.find_solutions(graph_interaction_settings_groups)
```

That can be wrapped in a simple function like `expertsystem.generate_transitions`. A facade like that that can then be used in:
- [ ] the visualization page (see e.g. #341)
- [ ] the interactive pages
- [ ] an 'actual' quickstart -- one that fits in one page, with references to the 'old' quickstart (now named something like "workflow") with more explanations.
- [ ] `tensorwaves` (see [Usage: Step 1](https://pwa.readthedocs.io/projects/tensorwaves/en/latest/usage/1_create_model.html) there)

See also #274

redeboer:
Closes #343 

Provides a few facades that simplify the use of the `expertsystem`. The usage main page has become a notebook where these facades are shown in a simple overview. The interactive page has also been expanded so that the `expertsystem` can be used as an online utilities.


spflueger:
This overlaps with  #281 now

Since the interface of the solver code change quite a bit, the code should be adapted to this a bit.

A small side remark, I would not combine the functionality of returning all intermediate states with the violations, but rather split this up into two functions. So this PR should concentrate on the violation finding.

redeboer:
Yeah we have to reconsider this a bit: what would an average website visitor want to see? For instance, in this "Check reaction" section, does it make sense to generate intermediate states, or would one want to check a specific reaction chain where one does know the intermediate states? And can we make it more general in usage?

Key part for such an interactive page is that it remains flexible but also contained within just a few lines.

redeboer:
Can move that yes, this PR has become a bit vague as long as #281 is unresolved.

PR #275, opened 2020-09-28
docs: improve docstring of Rule

redeboer:
An attempt to make clearer that the type hints of the `Rule` derivative classes have an important dynamic role as well.

### DIFF
--- a/expertsystem/solving/conservation_rules.py
+++ b/expertsystem/solving/conservation_rules.py
@@ -47,21 +47,24 @@ class Rule:
     instances. The `__call__` method contains actual rule logic and has to be
     overwritten.
 
+    .. warning::
+        Besides the rule logic itself, a `Rule` also has the responsibility of
+        stating its run conditions. These run conditions **must** be stated by
+        the type annotations of its `.__call__` method. The type annotations
+        therefore are not just there for *static* type checking: they also
+        carry more information about the rule that is extracted *dynamically*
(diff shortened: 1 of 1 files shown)


expertsystem/state/conservation_rules.py
```diff
@@ -47,21 +47,24 @@ class Rule:
...
     overwritten.
 
+    .. warning::
+        Besides the rule logic itself, a `Rule` also has the responsibility of
+        stating its run conditions. These run conditions **must** be stated by
+        the type annotations of its `.__call__` method. The type annotations
+        therefore are not just there for *static* type checking: they also
+        carry more information about the rule that is extracted *dynamically*
+        by the `.propagation` module.
+
```
spflueger, line 57: Is there a reason you moved this block above? I would keep things logically together so, this should go below the paragraph
> For additive quantum numbers, the decorator .....

redeboer: Yeah I wasn't sure here. The `__annotations__` refer to the `__call__` method in the previous paragraphs, that's why I put it higher up

spflueger: maybe move that additive quantum number paragraph to the end then?

redeboer: Ok moved to the end

spflueger (approved):

PR #278, opened 2020-09-29
refactor!: simplify Propagator/Solver interface

### LINKED ISSUE #43: Add type hints

### LINKED ISSUE #258: Refactor the Propagator interface
This following code snippet illustrates the new interface design

```python3
@dataclass(frozen=True)
class Result:
    solutions: List[StateTransitionGraph] = []
    violated_rules: List[Set[Rule]] = []
    # a set of rules defines a group which together violates all possible combinations that were processed
    # how can this Set[Rule] be determined? The solver has to manage that, by keeping some kind of statistic
    # For now this passed/called statistic can help with most cases. This can be improved later on, but it
    # should be written in a modular way, so that some better code can easily be added instead.


class Propagator(ABC):
    """Interface of a Propagator."""

    @abstractmethod
    def find_solutions(
        graph: StateTransitionGraph, settings: InteractionNodeSettings
    ) -> Result:
        pass
```

spflueger:
Closes #43 
Closes #258


redeboer (approved):
Good job! It's an intermediate state (that does not break things) so let's move on

PR #280, opened 2020-09-30
chore!: reorganize code for new subpackage solving

spflueger:


redeboer (approved):
Looks good. We may want to think about three submodules that follow the usual workflow:
- `particles`
- `solving`
- `amplitude`/`pwa`

Not sure what to do with `state` though. See #196.

PR #285, opened 2020-10-01
refactor!: change solutions type of Result class

### LINKED ISSUE #283: Use Particle with Spin projection in solutions of Result
- [ ] change Result to contain solutions of the type `StateTransitionGraph[ParticleWithSpin]` instead of `StateTransitionGraph[dict]`
- [ ] move `initialize_graphs_with_particles()` call into  `__run_csp()` and merge functionality with `__apply_solutions_to_graph()`

spflueger:
Closes #283 

Note that the Result is currently a `Generic` to allow the StateTransitionManager to return the old type of Result, while the solving module is using the new `ParticleWithSpin`.

### DIFF
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -149,6 +149,30 @@ for edge_qn_name, edge_qn_type in EdgeQuantumNumbers.__dict__.items():
         edge_qn_type.__module__ = "expertsystem.data"
 
 
+# for static typing
+EdgeQuantumNumber = Union[
+    EdgeQuantumNumbers.pid,
+    EdgeQuantumNumbers.mass,
+    EdgeQuantumNumbers.width,
+    EdgeQuantumNumbers.spin_magnitude,
+    EdgeQuantumNumbers.spin_projection,
+    EdgeQuantumNumbers.charge,
+    EdgeQuantumNumbers.isospin_magnitude,
+    EdgeQuantumNumbers.isospin_projection,
+    EdgeQuantumNumbers.strangeness,
+    EdgeQuantumNumbers.charmness,
+    EdgeQuantumNumbers.bottomness,
+    EdgeQuantumNumbers.topness,
+    EdgeQuantumNumbers.baryon_number,
+    EdgeQuantumNumbers.electron_lepton_number,
+    EdgeQuantumNumbers.muon_lepton_number,
+    EdgeQuantumNumbers.tau_lepton_number,
+    EdgeQuantumNumbers.parity,
+    EdgeQuantumNumbers.c_parity,
+    EdgeQuantumNumbers.g_parity,
+]
+
+
 @dataclass(frozen=True, init=False)
 class NodeQuantumNumbers:
     """Definition of quantum numbers for interaction nodes."""
@@ -164,6 +188,16 @@ for node_qn_name, node_qn_type in NodeQuantumNumbers.__dict__.items():
     if not node_qn_name.startswith("__"):
         node_qn_type.__qualname__ = f"NodeQuantumNumbers.{node_qn_name}"
         node_qn_type.__module__ = "expertsystem.data"
+
+
+# for static typing
+NodeQuantumNumber = Union[
+    NodeQuantumNumbers.l_magnitude,
+    NodeQuantumNumbers.l_projection,
+    NodeQuantumNumbers.s_magnitude,
+    NodeQuantumNumbers.s_projection,
+    NodeQuantumNumbers.parity_prefactor,
+]
 
 
 @dataclass(frozen=True)
--- a/expertsystem/nested_dicts.py
+++ b/expertsystem/nested_dicts.py
@@ -224,30 +224,30 @@ QNNameClassMapping = {
 }
 
 edge_qn_to_enum = {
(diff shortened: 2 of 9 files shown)


expertsystem/nested_dicts.py
```diff
@@ -309,3 +321,85 @@ def _particle_with_spin_projection_to_dict(
         if item["Type"] == "Spin":
             item["Projection"] = spin_projection
     return output
+
+
+def _convert_nodes_to_dict(
```
redeboer, line 326: It doesn't really matter while this function is temporary (right?), but `_convert_nodes_to_dict` modifies its input, while `_convert_edges_to_dict` creates a new object and returns it.

redeboer, line 326: It doesn't really matter while this function is temporary (right?), but `_convert_nodes_to_dict` modifies its input, while `_convert_edges_to_dict` creates a new object and returns it.

expertsystem/solving/__init__.py
```diff
@@ -223,25 +264,33 @@ def _extract_value(props: Dict[Enum, Any], obj_type: Any) -> Any:
 
 
 def _check_arg_requirements(
-    class_type: Type[Any], props: Dict[Enum, Any]
+    class_type: Type[Any], props: Dict[Enum, Any], key_mapping: Optional[dict]
```
redeboer, line 267: How permanent is this `key_mapping`?

expertsystem/solving/__init__.py
```diff
@@ -439,42 +499,105 @@ def _encode_variable_name(variable_info: _VariableInfo, delimiter: str) -> str:
...
 
 
+def _merge_solutions_with_graph(
+    solutions: List[QuantumNumberSolution],
+    graph: StateTransitionGraph[ParticleWithSpin],
+    allowed_particles: ParticleCollection,
+) -> List[StateTransitionGraph[ParticleWithSpin]]:
+    initialized_graphs = []
+
+    logging.debug("merging solutions with graph...")
```
redeboer, line 509: Is this an informative debugging statement?

expertsystem/solving/__init__.py
```diff
@@ -439,42 +499,105 @@
...
     def _create_edge_variables(
         edge_ids: Sequence[int],
-        qn_list: List[
-            Union[
-                ParticleDecayPropertyNames,
-                ParticlePropertyNames,
-                StateQuantumNumberNames,
-            ]
-        ],
+        qn_list: Set[Type[EdgeQuantumNumber]],
```
redeboer, line 600: Yes that's definitely more readible (and safer)

expertsystem/solving/__init__.py
```diff
@@ -439,42 +499,105 @@
...
+    }
+
+    if EdgeQuantumNumbers.isospin_magnitude in state:
+        changes_dict["isospin"] = Spin(
+            state[EdgeQuantumNumbers.isospin_magnitude],
+            state[EdgeQuantumNumbers.isospin_projection],
+        )
+    if EdgeQuantumNumbers.spin_magnitude in state:
+        changes_dict["spin"] = state[EdgeQuantumNumbers.spin_magnitude]
+    return replace(particle, **changes_dict) == particle
```
redeboer, line 578: Interesting, probably also good to use this in the implementation of `create_(anti_)particle`.

expertsystem/data.py
```diff
@@ -149,6 +149,30 @@ class EdgeQuantumNumbers:  # pylint: disable=too-many-instance-attributes
         edge_qn_type.__module__ = "expertsystem.data"
 
 
+# for static typing
+EdgeQuantumNumber = Union[
```
redeboer, line 153: We may want to split up this module into something like `types` (which is used internally) and `particles` (which is the 'starting point' interface of the `expertsystem`). See also #196 and also the new `amplitude.model` module (#287).

redeboer (approved):
Good job! I have a few thoughts/questions, but I approve because I know many issues resolve once #254 has been resolved.

expertsystem/nested_dicts.py
spflueger: Yeah, a function with a return would have been a better. But one of the follow up PRs will remove this code again anyway

expertsystem/solving/__init__.py
spflueger: Temporary

expertsystem/solving/__init__.py
spflueger: I thought its used there already, at least I remember I pointed this out before

expertsystem/solving/__init__.py
spflueger: True, I can remove this

expertsystem/solving/__init__.py
redeboer: Can be handled in #294

expertsystem/data.py
spflueger: Yes, we can address that in #196

expertsystem/solving/__init__.py
spflueger: I will remove it in one of the next PRs as this change is to superficial to rerun the CI for...

Issue #65, opened 2020-06-17
Modulate the amplitude module


spflueger:
- [x] Move write functionality to the `io` module: the `io` module just dumps those python classes to a file format of your choice.
- [x] Formulate the amplitude model in terms of python classes
- [x] (?) Remove `AbstractAmplitudeGenerator` in favour of functions, because the only remaining abstract method is `generate`.

redeboer:
I still don't think nested `dict`s are the way to go, especially after having gone through the [horror of creating a differently shaped nested dict from that](#59) and having to debug it.

Instead, I would prefer an object-oriented approach, such as a base class that represents a component and then derived from that an intensity/amplitude object that can then be further derived into `CoherentIntensity`, `IncoherentIntensity`, etc... Fixes the data throughout the framework, reveals intention (readability!), much easier to debug, and allows for tab completion (not unimportant for the DX).

spflueger:
> I still don't think nested `dict`s are the way to go, especially after having gone through the [horror of creating a differently shaped nested dict from that](#59) and having to debug it.
>
> Instead, I would prefer an object-oriented approach, such as a base class that represents a component and then derived from that an intensity/amplitude object that can then be further derived into `CoherentIntensity`, `IncoherentIntensity`, etc... Fixes the data throughout the framework, reveals intention (readability!), much easier to debug, and allows for tab completion (not unimportant for the DX).

I'm not sure we are talking about the same thing. I'm simply referring to the output to a file. I think all tools for various formats are working based on dicts (xmltodict, json, yaml..). So you are bound to use dicts at this point. Hence I simply want the amplitude generators to output a dict, which can be plugged into a tool of choice and dumped to file. I don't see how and OO class design is of any help here. How that dict is created from the solution, that's another topic.

redeboer:
It's still related, I think. What I mean is that the amplitude model should be created in an OO way (see [my description](https://github.com/ComPWA/expertsystem/issues/65#issuecomment-645464081)), while the output is separated from that. So only at the latest point (the output stage), there is an adapter that converts the amplitude model (which is built up of intensity objects) to a nested dictionary that can then be immediately dumped to YAML or whatever format. Then you have confined the representation to a separate place and you can change it at any stage without having to touch the amplitude model builder or whatever modules come before that.

[...]

PR #286, opened 2020-10-03
refactor!: move amplitude writers to io

redeboer:
In preparation to #65, writing functionality has been removed from the amplitude builders to the `io` module. Instead, they dump a `dict` (which will become formulated in terms of a dedicated `AmplitudeModel` class. So, just like with the `ParticleCollection`, the idea is that you use `io.write` to write an amplitude model to a file.


spflueger (approved):
Looks all right

Issue #134, opened 2020-07-29
Add entry point for fit parameters in the eventual amplitude recipe


redeboer:
See https://github.com/ComPWA/expertsystem/issues/127 and #124 

We want the `expertsystem` to construct possible resonances based on initial and final state and all states that are allowed in between according to conservation rules. These candidates are then converted into amplitudes (with fit parameters) in the eventual amplitude module.

Now, take resonance "mass" and "width" as an example. Ideally, we then want the `expertsystem` to suggest an 'optimal pole energy value' (https://github.com/ComPWA/expertsystem/issues/127#issuecomment-665740937) instead. However, see #72. So we need some user input at some stage.

redeboer:
Closed by #454

Issue #124, opened 2020-07-24
Add entry point where the user can specify decay dynamics


redeboer:
Since #189, the `Dynamics` section in the output amplitude model recipe file shrinks, because the PDG particle definitions do not contain dynamics info (e.g. Flatté or relativistic Breit-Wigner). This dynamics info should be added only later on for the eventual fitting package, so when the amplitude recipe is created.

We need to think of a smart design here, perhaps even leave it up to the fitting package.

redeboer:
@Leongrim How would you prefer this works form a user perspective?

Currently, you have to add your dynamics to the output recipe file and it should like this. The question is, how to make easily add this info through Python and at which stage (in terms of the workflow notebooks)?

https://github.com/ComPWA/tensorwaves/blob/d97bfdde2e1da1c6bb17f0d2d78c51a42d15f9ac/examples/intensity-recipe.yaml#L1195-L1224

Leongrim:
So frist of all. Anything you have to add yourselve should be avoided. But I think you now that 😁

redeboer:
That's true, but the `expertsystem` cannot provide dynamics, because it completely depends on your analysis.

To put it in other words: the `expertsystem` only knows initial state, final state, decay topology model (isobar), and conservation rules. Only at the moment of [amplitude model construction](https://pwa.readthedocs.io/projects/expertsystem/en/0.2.0-alpha/usage/quickstart.html#1.4.-Write-amplitude-model-to-recipe-file) does PWA start to come into the picture. And it's there that the dynamics start to play a part. (Dynamics basically just represent the fit function you use eventually in the fitter package, so that's technically separate from the allowed decay topologies that the `expertsystem` suggests.)

So for the dynamics, the question is, how to make it as convenient as possible to **insert** these dynamics to the amplitude construction?

Leongrim:
I think so smothest way would be to still let the expertsystem handle this an give the option to add the dynamics because otherwise youhad again the same functions in ComPWA and Tensorwaves.

So for the user I think it would be best to have a functionthats adds basic dynamics (e.g. breitwiegner) to all resonances and after that an easy way to change this dynamics inside of the fitting/simulation scripts.

So I dount know much about K-Matrix but does that not combine different final states that could bossible have the same resonances? If so it would also be good to add the dynamics to a transition and not to a resonance

redeboer:
This issue has become more pressing since https://github.com/ComPWA/tensorwaves/pull/120

[...]

PR #287, opened 2020-10-03
feat!: implement AmplitudeModel class

### LINKED ISSUE #65: Modulate the amplitude module

redeboer:
Closes #65 

The amplitude builders now generate a dedicated `AmplitudeModel` from list of `StateTransitionGraph`s. It can be navigate through purely in terms of python without making use of nested internal Python types. The model has therefore become independent of the nested XML `dict` structures and is well typed.

Note that this PR also partially addresses #124 and #134, however, this functionality is not tested and the `io.write` method cannot handle certain dynamics yet. This is easily fixed, however, and would be more strongly enforced through the tests written for #288.

_Rebased on #286_


redeboer:
@spflueger I can further split up this PR into:

- [ ] Removal of the abstract amplitude generator
- [ ] Smaller fixes
- [ ] Introduction of the amplitude interfaces

However, these steps only make sense when considering the whole (as is presented currently).

expertsystem/amplitude/model.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+@dataclass
+class FitParameter(abc.Hashable):
+    name: str
+    value: float = 0.0
+    is_fixed: bool = False
+
+    def __hash__(self) -> int:
+        return hash(str(self))
```
spflueger, line 30: there is a `hash=True` option in the `dataclass` decorator. Why is hashable required in the first place?

expertsystem/amplitude/model.py
```diff
@@ -0,0 +1,349 @@
...
+        return FitParameters(
+            {parameter for parameter in self.values() if function(parameter)}
+        )
+
+    def register_parameter(
+        self, name: str, value: float, fix: bool
+    ) -> FitParameter:
+        parameter = FitParameter(name=name, value=value, is_fixed=fix)
+        self.add(parameter)
+        return parameter
```
spflueger, line 84: I would remove that functionality as its not needed. So just use `add(FitParameter(...))`

expertsystem/amplitude/model.py
```diff
@@ -0,0 +1,349 @@
...
+    amplitudes: List[AmplitudeNode] = field(default_factory=list)
+
+
+@dataclass
+class CoefficientAmplitude(AmplitudeNode):
+    component: str
+    magnitude: FitParameter
+    phase: FitParameter
+    amplitude: AmplitudeNode
+    prefactor: Optional[FitParameter] = None
```
spflueger, line 267: The prefactor should not be a `FitParameter` but a constant, so `float`

tests/unit/amplitude/test_parity_prefactor.py
```diff
@@ -97,40 +88,11 @@ def test_parity_prefactor(
...
-
-    return None
+    prefactors = amplitude_model.parameters.filter(
+        lambda p: p.name.startswith("PreFactor")
+    )
+    if len(prefactors) > 0:
+        product = 1.0
+        for factor in prefactors.values():
+            product *= factor.value
+        assert relative_parity_prefactor == product
```
spflueger, line 98: I can't comprehend if this test is doing what it should be. So two very specific Coefficient amplitudes are supposed to have a this prefactor relation with each other. This code seems to check something else, as it just takes all defined prefactors and just multiplies up those values... Maybe I misunderstood something

spflueger (changes requested):
Good work. I just have a few questions

expertsystem/amplitude/model.py
redeboer: Problem is that you usually want to also register the parameter. So then you would always have to do:
```python
# parameters: FitParameters
par = FitParameter(...)
parameters.add(par)
```
So you can also turn it around: why make the `add` method public?

tests/unit/amplitude/test_parity_prefactor.py
redeboer: Actually I couldn't figure out what the existing test was doing. But I did notice that only one of the two cases actually had a PreFactor, so the other was just testing 1 = 1*1 because:
https://github.com/ComPWA/expertsystem/blob/e6471989258071ab2480fbf6b235c5a03bab05fb/tests/unit/amplitude/test_parity_prefactor.py#L117-L119

breaks the search loop upon finding the first `"Component"`.

However, note that e58141f2 does not break the test even though it should. So something is also going wrong in this test.

expertsystem/amplitude/model.py
redeboer: Ah yes 8687a6e5

expertsystem/amplitude/model.py
redeboer: e58141f2

tests/unit/amplitude/test_parity_prefactor.py
redeboer: Removed `register_parameter` 84eebf79

expertsystem/amplitude/model.py
redeboer: Removed `register_parameter` 84eebf79

tests/unit/amplitude/test_parity_prefactor.py
redeboer: Old test behaviour is back edeeb6d5

spflueger (approved):
GJ

PR #289, opened 2020-10-03
ci: ignore draft pull requests

redeboer:
Only run GitHub Actions if PR is ready for review (not draft status).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,8 +82,6 @@ jobs:
         run: make html
       - name: Test doctests in docstrings
         working-directory: doc
-        env:
-          NBSPHINX_EXECUTE: YES
         run: make doctest
 
   style:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ on:
     branches: [master]
   pull_request:
     branches: [master]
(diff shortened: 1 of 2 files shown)


redeboer:
Ok, works as expected:
- When in draft status, only PR linting and RTD is booted.
- Once marked ready for review, the rest of the checks are booted immediately, *without having to re-push*

.github/workflows/ci.yml
```diff
@@ -81,8 +82,6 @@ jobs:
         run: make html
       - name: Test doctests in docstrings
         working-directory: doc
-        env:
-          NBSPHINX_EXECUTE: YES
```
spflueger, line 85: why can this be removed now?

redeboer: The doctest job should only check of `doctest` directives in the docstrings. Now it also runs the notebooks, which makes the workflow longer by some two minutes.

Note that the notebooks are already run in the job prior to it:
https://github.com/ComPWA/expertsystem/pull/289/files/7d18e93af3011db9576dc56a5a388b2e59e86400#diff-e9f950f17198d3d5e3122a44230a09b9R78-R82

spflueger: i see

spflueger (approved):

PR #291, opened 2020-10-05
feat!: implement load_amplitude_model

### LINKED ISSUE #288: Load AmplitudeModel from file formats
Since #287, the `io` module can write an `AmplitudeModel` instance, but it cannot load one from YAML or XML. Once this is implemented, the `AmplitudeModel` can be serialized back and forth. 

If an XML loader is also implemented, the amplitude writer tests can be made more reliable: comparisons between the two file formats have previously highlighted several bugs in `ParticleCollection` and, related, the `particle` package.

redeboer:
Closes #288

### DIFF
--- a/expertsystem/io/yaml/_build.py
+++ b/expertsystem/io/yaml/_build.py
@@ -61,6 +61,63 @@ def build_amplitude_model(definition: dict) -> AmplitudeModel:
     )
 
 
+def build_particle_collection(
+    definition: dict, validate: bool = True
+) -> ParticleCollection:
+    if validate:
+        validation.particle_list(definition)
+    definition = definition["ParticleList"]
+    particles = ParticleCollection()
+    for name, particle_def in definition.items():
+        particles.add(build_particle(name, particle_def))
+    return particles
+
+
+def build_particle(name: str, definition: dict) -> Particle:
+    qn_def = definition["QuantumNumbers"]
+    return Particle(
+        name=name,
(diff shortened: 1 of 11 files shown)


expertsystem/io/yaml/_build.py
```diff
@@ -15,8 +41,206 @@
...
+    return AmplitudeModel(
+        particles=particles,
+        kinematics=kinematics,
+        parameters=parameters,
+        intensity=intensity,
+        dynamics=dynamics,
+    )
+
+
+def build_fit_parameters(definition: List[dict]) -> FitParameters:
```
spflueger, line 62: Did you leave all of those functions, which "build" a portion of the full AmplitudeModel public on purpose?

redeboer: Yes it's used here
https://github.com/ComPWA/expertsystem/blob/ef358578b7583a79d1fe05110fd4ea727f0a0595/expertsystem/io/yaml/_build.py#L47

Or do you mean, why is it public?

spflueger: yes why is it public, and all of the functions except `build_amplitude_model`

redeboer: Ok c442ef3c ff8f6ba5 40dc58a0 makes the implementations private. We should think about merging these dumper modules though because of the many duplicate lines. For a follow-up issue.

spflueger (approved):

PR #292, opened 2020-10-05
refactor: change ParticleCollection into set

redeboer:
This makes the `ParticleCollection` behave a bit more like expected. So for instance no need to use `keys`, `values`, and `items` when using over a `ParticleCollection` instance.

### DIFF
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -65,9 +65,14 @@ class Spin(abc.Hashable):
                 f"Spin magnitude {magnitude} has to be a multitude of 0.5"
             )
         if abs(projection) > magnitude:
+            if magnitude < 0.0:
+                raise ValueError(
+                    "Spin magnitude has to be positive:\n" f" {magnitude}"
+                )
             raise ValueError(
-                "Spin projection cannot be larger than its magnitude:\n"
-                f"  {projection} > {magnitude}"
+                "Absolute value of spin projection cannot be larger than its "
+                "magnitude:\n"
+                f" abs({projection}) > {magnitude}"
             )
         if not (projection - magnitude).is_integer():
             raise ValueError(
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -320,8 +320,16 @@ class ParticleCollection(abc.Set):
                 }
             )
 
-    def __contains__(self, particle_name: object) -> bool:
-        return particle_name in self.__particles
+    def __contains__(self, instance: object) -> bool:
+        if isinstance(instance, str):
+            return instance in self.__particles
+        if isinstance(instance, Particle):
+            return instance in self.__particles.values()
+        if isinstance(instance, int):
+            return instance in {p.pid for p in self}
+        raise NotImplementedError(
+            f"Cannot search for type {instance.__class__.__name__}"
+        )
 
     def __eq__(self, other: object) -> bool:
         if isinstance(other, abc.Iterable):
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -520,7 +520,7 @@ def create_particle(  # pylint: disable=too-many-arguments,too-many-locals
     return Particle(
         name=name if name else template_particle.name,
         pid=pid if pid else template_particle.pid,
-        mass=mass if mass else template_particle.mass,
+        mass=mass if mass is not None else template_particle.mass,
         width=width if width else template_particle.width,
         spin=spin if spin else template_particle.spin,
         charge=charge if charge else template_particle.charge,
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -367,7 +367,7 @@ class ParticleCollection(abc.MutableSet):
         if isinstance(other, Particle):
             self.add(other)
         elif isinstance(other, ParticleCollection):
-            self.merge(other)
+            self.update(other)
         else:
             raise NotImplementedError(f"Cannot add {other.__class__.__name__}")
         return self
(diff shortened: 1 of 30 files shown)


expertsystem/data.py
```diff
@@ -375,7 +390,7 @@ def find(self, search_term: Union[int, str]) -> Particle:
         if isinstance(search_term, int):
             pid = search_term
             search_results = [
-                particle for particle in self.values() if particle.pid == pid
+                particle for particle in self if particle.pid == pid
```
spflueger, line 393: I'm thinking if this `find` can be combined with `filter`.

expertsystem/data.py
```diff
@@ -320,8 +320,23 @@ def __init__(self, particles: Optional[Iterable[Particle]] = None) -> None:
                 }
             )
 
-    def __contains__(self, particle_name: object) -> bool:
-        return particle_name in self.__particles
+    def __contains__(self, instance: object) -> bool:
+        if isinstance(instance, str):
+            return instance in self.__particles
+        if isinstance(instance, Particle):
+            return instance in self.__particles.values()
```
spflueger, line 327: This could be `return instance.name in self.__particles`

expertsystem/data.py
```diff
@@ -301,13 +301,13 @@ def compute_isospin_projection(  # pylint: disable=too-many-arguments
         )
 
 
-class ParticleCollection(abc.Mapping):
-    """Safe, `dict`-like collection of `.Particle` instances."""
+class ParticleCollection(abc.Set):
```
spflueger, line 304: Shouldn't it be a `MutableSet`. I think support for the deletion or update of elements would also be convenient (currently there is only merge).

redeboer: Yes that is true. And then probably the `add` method makes more sense as well (it's in the `MutableSet` base class).

expertsystem/data.py
redeboer: No, because it checks whether the `Particle` instance is in the `ParticleCollection` 'set'. It's also tested here
https://github.com/ComPWA/expertsystem/blob/f2323fe1bd079ea41b37a2565bc431ba36d0be67/tests/unit/test_data.py#L269

expertsystem/data.py
redeboer: I prefer not, because `find` returns a `Particle` instance whereas `filter` returns a `ParticleCollection`.

spflueger: I know, but I think that definition of `find` is confusing as it will raise an exception if more than one candidate is found. I would expect to get all results or the first occurrence. I prefer all results. The return type can also `Union[Particle, List[Particle]]` for convenience.

expertsystem/data.py
redeboer: 30de4cdc and a41bb834

expertsystem/data.py
redeboer: I greatly to keep the return type unambiguous. But yes, the exception for duplicate PIDs is dubious. Actually that should already be checked when adding, because we want the PIDs to be unique (for instance because it's used as a check in the conservation rules).

expertsystem/data.py
spflueger: Well its the same thing (just faster), as you guarantee that the string keys are unique and match their particles.

This actually touches an issue. What if you define a "faulty" particle with different properties, but the same name as an existing particle in the collection. Should that return a `True` or a `False` for `__contains__`?

redeboer: That is true. I had a different check in mind, so maybe you're right. But technically, for a `set` only `instance in self.__particles.values()` should be the correct behaviour.

expertsystem/data.py
spflueger: > Actually that should already be checked when adding, because we want the PIDs to be unique (for instance because it's used as a check in the conservation rules).

Right. It should actually be the other way around, that searching using a string can return multiple candidates.

expertsystem/data.py
redeboer: With #295 implemented, is the existing implementation correct?

expertsystem/data.py
redeboer: I'm still a bit hesitant to touch the `find` method. Currently, it is guaranteed that the names unique (if not, the instance is overwritten and a warning is printed). For PID there is now also such a warning, but there can still exist a situation where there are two equivalent particle with the same PID. That's why an exception is raised there.

spflueger: I think now with #295 this makes more sense as the str and pid are more sorting attributes of the `ParticleCollection` class now. The pid should also be handled in a unique way by the ParticleCollection.

expertsystem/data.py
spflueger: I would just make sure that the pid is also assigned uniquely. A second mapping pid <-> Particle can be used similar to the str one.

redeboer: Ok is checked now 3951fd64. No test the logging though, wonder if it should raise a `UserWarning` or so

expertsystem/data.py
redeboer: So PID and name are handled on an equal footing now, I think 3951fd64

expertsystem/data.py
```diff
@@ -343,31 +343,29 @@ def compute_isospin_projection(  # pylint: disable=too-many-arguments
...
-        return particle_name in self.__particles
+            self.update(particles)
+
+    def __contains__(self, instance: object) -> bool:
+        if isinstance(instance, str):
+            return instance in self.__particles
+        if isinstance(instance, Particle):
+            return instance in self.__particles.values()
+        if isinstance(instance, int):
+            return instance in {p.pid for p in self}
```
spflueger, line 361: this would change to `instance in self.__pid_to_name`

redeboer: Oops, yes 14cf964f

spflueger (approved):
gj

PR #295, opened 2020-10-06
feat: ignore name and pid in Particle comparison

### LINKED ISSUE #293: Ignore name and pid in Particle comparison
The `name` and `pid` attributes of `Particle` should become mere labels so that only the quantum numbers and properties like mass and width make a `Particle` uniquely distinguishable. This also has consequences for the behaviour of `ParticleCollection`. For more info see the comments in PR #292.

redeboer:
Closes #293

### DIFF
--- a/expertsystem/data.py
+++ b/expertsystem/data.py
@@ -2,7 +2,7 @@
 
 import logging
 from collections import abc
-from dataclasses import dataclass, fields
+from dataclasses import dataclass, field, fields
 from functools import total_ordering
 from typing import (
     Any,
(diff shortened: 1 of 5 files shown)


redeboer:
@spflueger This PR is currently rather trivial, because I thought it would be more complicated. But maybe I'm missing something. For instance, how should the `__hash__` method of a `Particle` behave now?

spflueger (approved):

spflueger:
> @spflueger This PR is currently rather trivial, because I thought it would be more complicated. But maybe I'm missing something. For instance, how should the `__hash__` method of a `Particle` behave now?

Things should work as expected (see note). You can add a test for checking the hash is the same for the two particles with different name or pid

```python3
assert hash(particle) == hash(different_name_particle)
```

Note: Afaik the dataclasses module will create a `__hash__` method now (since particle if frozen and comparable). As the name and pid are stated as non comparable fields, these values should neither be in the `__hash__` implementation nor the `__eq__` (generated by dataclasses).

redeboer:
> Afaik the dataclasses module will create a __hash__ method now (since particle if frozen and comparable).

Yes, so actually this needs to be fixed still (and indeed tested), because the `__hash__` method is currently overwritten.

spflueger (changes requested):
Ah I missed that. The `__hash__` function should be removed

redeboer:
> Ah I missed that. The __hash__ function should be removed

Ok done now. There are still some things to be fixed now in the `ParticleCollection` but better to address that in #292

spflueger (approved):

PR #297, opened 2020-10-06
refactor!: fully abolish old nested dict graph in STM

### LINKED ISSUE #131: Use StateTransitionGraphs with Particles in solver code
- [x] change solver interface to from `StateTransitionGraph[dict]` to `StateTransitionGraph[ParticleWithSpin]` 
- [x] replace `get_particle_property()` function calls with new code that extract the info from the `Particle` + spin projection tuple

### LINKED ISSUE #142: Remove get_particle_property function
The `io.xml` module already provides this functionality (the XML particle properties are just baked into `Particle`). It will require some refactoring, but this function can be removed.

Note that `get_interaction_property` is slightly different, see #141

spflueger:
Closes #131
Closes #142

### DIFF
--- a/expertsystem/topology.py
+++ b/expertsystem/topology.py
@@ -316,7 +316,7 @@ class StateTransitionGraph(Topology, Generic[_EdgeType]):
         super().__init__(nodes, edges)
         self.node_props: Dict[int, dict] = {}
         self.edge_props: Dict[int, _EdgeType] = {}
-        self.graph_element_properties_comparator: Optional[Callable] = None
+        self.graph_node_properties_comparator: Optional[Callable] = None
 
     def __repr__(self) -> str:
         return (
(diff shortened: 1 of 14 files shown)


expertsystem/topology.py
```diff
@@ -316,7 +316,7 @@ def __init__(
         super().__init__(nodes, edges)
         self.node_props: Dict[int, dict] = {}
         self.edge_props: Dict[int, _EdgeType] = {}
-        self.graph_element_properties_comparator: Optional[Callable] = None
+        self.graph_node_properties_comparator: Optional[Callable] = None
```
redeboer, line 319: Does the caller always have to specify the comparator? Maybe better to provide a default one?

expertsystem/ui/__init__.py
```diff
@@ -378,14 +376,12 @@ def find_solutions(
             )
 
         # merge strengths
-        final_result = Result[ParticleWithSpin]([])
+        final_result = Result([])
```
redeboer, line 379: Still don't really like that the caller has the responsibility of initializing an empty list for the `Result.solutions`.

redeboer (approved):

expertsystem/ui/__init__.py
spflueger: Thats true

expertsystem/topology.py
spflueger: 2d633a37a070d3397fb0bcddf633f28acb7239cc

expertsystem/ui/__init__.py
spflueger: 2d633a37a070d3397fb0bcddf633f28acb7239cc

redeboer (approved):

PR #307, opened 2020-10-08
style!: enforce isort sorting

redeboer:
Enforced through both VSCode format-on-save and pre-commit.


redeboer:
@spflueger better to merge #301 first due to the eventual merge conflicts

spflueger (approved):

PR #308, opened 2020-10-09
chore: remove numpy as a dependency

### LINKED ISSUE #306: Remove numpy as a dependency
By now, `numpy` is only used for `arange` and `sign`.

See https://github.com/ComPWA/expertsystem/pull/301#discussion_r501798816

redeboer:
Closes #306 

A benefit is for instance that this reduces the risk of dependency issues like #241

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -69,7 +69,6 @@
         "mathbb",
         "mypy",
         "nishijima",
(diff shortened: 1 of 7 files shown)


spflueger (approved):
Good job

PR #312, opened 2020-10-09
refactor!: replace dataclasses with attrs

### LINKED ISSUE #198: Use attrs package instead of dataclass
Since #135 we dropped support for Python 3.6. This can be avoided by using the [`attrs` package](https://www.attrs.org/). It possibly even improves the inheritance structures we use (`dataclass` does not give control over attribute construction order of derived classes).

redeboer:
Closes #198


spflueger (approved):
LGTM

PR #313, opened 2020-10-09
refactor!: remove nested dict graph node props

### LINKED ISSUE #141: Create node properties class to avoid nested dicts in graph
Related to #142, but more complicated due to #127

spflueger:
Closes #141


redeboer (approved):
I like it, good job :+1:

Issue #265, opened 2020-09-23
Reduce respontibilities of the StateTransitionManager


redeboer:
The `StateTransitionManager` is supposed to serve as a facade to the `expertsystem`, but many of its methods (such as `_determine_node_settings`) perform quite a lot of tasks. That raises the question whether they shouldn't become functions elsewhere, so that the `StateTransitionManager` can be made smaller.

PR #314, opened 2020-10-12
refactor!: reorganise submodules

### LINKED ISSUE #196: Reconsider submodule structure

### LINKED ISSUE #305: Isolate definitions related to solving.types submodule
Currently, classes such as `EdgeQuantumNumbers`, which are used from typing, are defined in the `particles` module. However, they are for internal use only, so it's better to have them placed under a separate module.

redeboer:
Closes #305 and closes #196 

The `expertsystem` now consists of three main components that form a workflow sequence:
1. `particle`, which is just about loading and filtering quantum numbers (`Particle` and related classes)
2. `reaction` ― the core of the `expertsystem` that generates allowed interactions as a `list` of `StateTransitionGraph`s
3. `amplitude`, where the `StateTransitionGraph` solutions are converted into amplitude models according to PWA formalisms.

For the upcoming release, we leave `ui` and `io` as they are, but we should eventually move more of their functionality into these three main components (`reaction` in particular). The main `__init__.py` files of `expertsystem`, `particle`, `reaction`, and `amplitude` should then become the User Interfaces of their respective submodules (that is, containing facade functions to their submodules). For instance, after the upcoming release, we will reduce the number of responsibilities of `StateTransitionManager` by moving its methods into the submodules of `reaction` and just having one `prepare_graph` function (not a method) in the `reaction/__init__.py` (see #265).

Some other considerations:
* `topology`: even though this module formulates the `StateTransitionGraph` and related functionality, we will still name it "`topology`" (not `graph`) because the "decay topology" is a common expression.
* `properties` has been split up and moved into `ui`, `combinatorics`, and the `reaction.__init__.py`
* Avoid general names such as `utils` and `types`. So `utils` is gone and `types` has become `quantum_numbers`.


expertsystem/types.py
```diff
@@ -0,0 +1,137 @@
+"""Definitions used internally for typing and signatures."""
```
spflueger: These are all solving related types (except Scalar maybe). I would propose to move this into the solving module, because solving actually defines these types

expertsystem/state/__init__.py
```diff
@@ -1,8 +0,0 @@
-"""All modules related to a quantum mechanical state."""
-
-__all__ = [
-    "combinatorics",
-    "properties",
-]
-
-from . import combinatorics, properties
```
spflueger, line 8: I just comment here, since i can't find a better place. Is it possible to abolish this properties module and move the functionality to appropriate modules? Things that have to do with solving can go into solving modules. Things that have to do with ui more into system control or so.

Mainly I want to have the code logically grouped together

expertsystem/amplitude/canonical_decay.py
```diff
@@ -4,7 +4,7 @@
 
 from expertsystem.data import Spin
 from expertsystem.solving.topology import StateTransitionGraph
-from expertsystem.types import InteractionProperties, ParticleWithSpin
+from expertsystem.solving.types import InteractionProperties, ParticleWithSpin
```
spflueger, line 7: Ok good, so my first comment is already addressed. I personally don't like the name types very much. Maybe rename it into `quantum_number` or something more meaningful?

expertsystem/__init__.py
```diff
@@ -17,11 +17,11 @@
...
-    "data",
+    "particles",
     "io",
     "solving",
     "ui",
 ]
 
 
-from . import amplitude, data, io, solving, ui
+from . import amplitude, io, particles, solving, ui
```
spflueger, line 39: wouldn't singular form be more matching? so `particle`?

spflueger (commented):
Looks good! I just have a few small comments

redeboer:
Update:
Feedback has been addressed in team discussion and follow-up commits. A summary can be founding **updated PR description**.

redeboer:
The `solving` module has been renamed into `reaction` now (see updated PR description). The original `expertsystem.solving` module has become `expertsystem.reaction.solving` without submodules.

expertsystem/solving/quantum_numbers.py
```diff
@@ -0,0 +1,144 @@
...
+
+from typing import Dict, NewType, Optional, Tuple, Type, Union
+
+import attr
+
+from expertsystem.particles import Parity, Particle
+
+Scalar = Union[int, float]
+
+ParticleWithSpin = Tuple[Particle, float]
```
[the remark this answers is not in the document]
redeboer: So I left `ParticleWithSpin` in for now, the rest has been moved to `reaction.solving`.

spflueger (approved):
Good job

PR #315, opened 2020-10-13
fix: avoid deep copies of graphs during solving

### LINKED ISSUE #310: Avoid deepcopy of StateTransitionGraph
We currently use `deepcopy` to clone a `StateTransitionGraph`, but this creates several identical `Particle` instances.

spflueger:
Closes #310


redeboer (approved):
Looks good! And indeed a reduction in CI time :+1: 
Triggered by 1392899, I opened #316

PR #317, opened 2020-10-13
ci: report and clean up unused code

### LINKED ISSUE #316: Report unused code in CI
Since `pyright` is run in the CI now (#247), unused variables, class definitions etc, can be reported automatically, see
https://github.com/microsoft/pyright/blob/master/docs/configuration.md

Easy to activate this. However, as the follow-up will show, we will need to do some refactoring to fix this dead code. (Therefore a separate issue for it.)

redeboer:
Closes #316


redeboer:
@spflueger The CI will fail for now, because we need to do some cleanup. Hence a question label.

spflueger (approved):
Nice addition!

spflueger:
So I fixed the unused code pieces you mentioned. 

> So, afaics, we need to think about the following unused code:
> https://github.com/ComPWA/expertsystem/blob/c4f1769c7c49915ca363b8402e82826b3b30c922/expertsystem/reaction/conservation_rules.py#L753
> 
> 
> Not used at all

Simply can be removed, was just an copy and paste error...

> https://github.com/ComPWA/expertsystem/blob/c4f1769c7c49915ca363b8402e82826b3b30c922/expertsystem/ui/_system_control.py#L58-L62
> 
> 
> Only used in `test_y_to_d0_d0bar_pi0_pi0.py`

Was only introduced in that specific test to speed up the test and not test too many combinations. So in other words a speed workaround.... With the improvements of splitting up the magnitude and projection quantum numbers of a spin. The combinatorics is kept at a minimum, so that this can also be just removed for now.

A feature to change the default settings or to flexibly create new settings would be needed at some point though.

> https://github.com/ComPWA/expertsystem/blob/c4f1769c7c49915ca363b8402e82826b3b30c922/expertsystem/ui/_system_control.py#L80-L82
> 
> 
> Only used in `test_ls_coupling.py`.
> @spflueger
> This is what `pyright` fishes out, so I would focus only on what to do with these. There's also some code that can be merged as well, but that's a broader issue.

Reformulated the ls coupling test, as this was currently not a unit test. Now its actually just testing the spin coupling and testing various Clebsch Gordan edge cases

PR #320, opened 2020-10-15
fix: validate list-only in recoil systems

redeboer:
There's a mistake in the JSON validation schema for the YAML recipe file: it does not reflect the structure of a `RecoilSystem`, which has a `list` of `int`s only:
https://github.com/ComPWA/expertsystem/blob/c4f1769c7c49915ca363b8402e82826b3b30c922/expertsystem/amplitude/model.py#L329-L332

### DIFF
--- a/expertsystem/schemas/yaml/amplitude-model.json
+++ b/expertsystem/schemas/yaml/amplitude-model.json
@@ -313,22 +313,12 @@
             "additionalProperties": false,
             "properties": {
               "RecoilFinalState": {
-                "anyOf": [
-                  { "$ref": "#/definitions/StateID" },
-                  {
-                    "type": "array",
-                    "items": { "$ref": "#/definitions/StateID" }
-                  }
-                ]
+                "type": "array",
+                "items": { "$ref": "#/definitions/StateID" }
               },
               "ParentRecoilFinalState": {
-                "anyOf": [
-                  { "$ref": "#/definitions/StateID" },
-                  {
-                    "type": "array",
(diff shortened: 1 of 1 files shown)


spflueger:
In the original xml file, both an array of IDs as well as a single ID was valid. How is that handled now?

redeboer:
https://github.com/ComPWA/expertsystem/pull/320#issuecomment-709416611
An alternative solution would be to modify the dump/builders

spflueger:
I see, so the dumpers and parsers work only with arrays?

Even though `[2]` is a bit more verbose than `2`, its more clear what is actually expected there. So I would say thats all right. What do you think?

redeboer:
> I see, so the dumpers and parsers work only with arrays?
> 
> Even though `[2]` is a bit more verbose than `2`, its more clear what is actually expected there. So I would say thats all right. What do you think?

I agree. It should be fixed now, however, I realise this problem wasn't identified because (1) there is no test on a decay to 4 particles (so we only test `"RecoilFinalState"`) and (2) the dumper always dumps to `List[int]`, so reading back from that (which the tests do) doesn't test this `anyOf`.

redeboer:
> (2) the dumper always dumps to `List[int]`, so reading back from that (which the tests do) doesn't test this `anyOf`.

Should be addressed with fa181620 now

redeboer:
Ok so I misunderstood slightly. Now the PR has been reverted to the original commit 5f49fdf7 (rebased)

spflueger (approved):

PR #323, opened 2020-10-16
fix: remove conda-forge channel

redeboer:
See https://github.com/ComPWA/tensorwaves/pull/133

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,6 +1,5 @@
 name: es
 channels:
-  - conda-forge
   - defaults
 dependencies:
   - python>=3.6


spflueger:
So something was missing in the previous PR?

redeboer:
> So something was missing in the previous PR?

Wow this is weird yes. Sorrry I messed up something extracting this commit from #324

spflueger (approved):

redeboer:
@spflueger Now I think it may make sense to make one job that checks the Conda developer setup. This problem arose because 
 conda-forge made 3.9 available and I only noticed when I recreated my environment. There may be other smaller issues when running conda, so might be a good idea to have that tested.

I'll push a try-out commit, then you can judge.

spflueger:
Sounds good

redeboer:
Hmm, I give up. To get it working you have to apply so many tricks that I doubt whether it's still a proper 'test'. So I'll leave it with the bug fix (will merge once CI is done)

PR #324, opened 2020-10-16
docs: visualize module dependencies

redeboer:
Also adds links to related projects in the sidebar and hides the ADR pages.

Note: `pydeps` requires Graphviz at runtime. However, the `subprocess.call` won't signal a fatal error once `pydeps` fails and then the image simply won't be added. On RTD, Graphviz is available (see the graph vizualisation module), and the image will be rendered if it is available.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -70,6 +70,7 @@
         "mypy",
         "nishijima",
         "pycompwa",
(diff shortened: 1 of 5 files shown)


spflueger (approved):
Nice

PR #326, opened 2020-10-16
refactor!: merge ui into reaction and amplitude

### LINKED ISSUE #319: Merge ui submodule into expertsystem.__init__.py

redeboer:
Closes #319


expertsystem/reaction/solving.py
```diff
@@ -91,12 +91,14 @@ def __init__(
         ] = None,
         not_executed_rules: Optional[Dict[int, Set[Rule]]] = None,
         violated_rules: Optional[Dict[int, Set[Tuple[Rule]]]] = None,
+        formalism_type: Optional[str] = None,
```
spflueger, line 94: Why is that optional? Shouldn't that be always given?

redeboer: The `Result` is used in the `solving` module, where the `formalism_type` string is not known (it is only known to the STM).

spflueger: Right. Ok thats def a code smell. We can leave it like this for now, but in the future this has to be handled better. It suggests that a composite type Result + formalism_type is needed. Maybe the root issue lies somewhere else that we have to refactor at that point.

spflueger (approved):

Issue #325, opened 2020-10-16
Formulate List[StateTransitionGraph] in terms of a class


redeboer:
I think it would be safer to wrap the allowed transitions (currently represent by a `list` of `StateTransitionGraph`s) into a class, something like `StateTransitionCollection`. This object can then contain more advanced methods to give info about the `list` of `StateTransitionGraph`s it contains, as well as some checks when adding new `StateTransitionGraph`s (for instance to enforce that the initial state and final state particles are consistent).

redeboer:
@spflueger moving the discussion of https://github.com/ComPWA/expertsystem/pull/328#pullrequestreview-512083599 here for now

The `Result` class currently provides:
1. `solutions` (`List[StateTransitionGraph[ParticleWithSpin]]`)
2. `violated_rules`
3. `not_executed_rules`
4. `formalism_type` (fishy that this would be necessary when computing allowed transitions, but this is another discussion)

If I understand correctly, only 1. and 4. are useful once there are solutions, while 2. and 3. are of use only when there are no solutions. To me that suggests it's better that what is currently the `Result` class becomes a class as suggested in this issue (a wrapper around `List[StateTransitionGraph]`) and that we raise a dedicated exception class (`NoSolutionsFound` or something) that carries `violated_rules` and `not_executed_rules`.

redeboer:
The `formalism_type` can still be contained within the returned class in case there are solutions, however, I still prefer if we get rid of `formalism_type` in the `reaction` module eventually.

spflueger:
> The `formalism_type` can still be contained within the returned class in case there are solutions, however, I still prefer if we get rid of `formalism_type` in the `reaction` module eventually.

I thought about this a bit more. I think the only thing that makes this difficult is that the spin projections have different meanings in the different formalisms. So in the helicity formalism its helicities, while in the canonical its spin projections onto some fixed axis.

To generate a result that is "independent" of the formalism, which could be put into any of the amplitude builders. I currently do not see a way around this other than computing the spin projections in parallel. Or some kind of adapter would be needed to convert.

The good thing is that we currently do not support the canonical formalism natively, but put the canonical formalism on top of the helicity formalism (substituting helicity amps with a sum of canonical ones). Hence we only have the helicity spin projections and do not run into the problem above. Long term this problem will arise though.

spflueger:
[...]

PR #328, opened 2020-10-16
feat: move getter methods to Result

redeboer:
Moves the `get_intermediate_state_names` function into `Result`. Also defined related function `Result.get_initial_state` and `Result.get_final_state` and implements them in the `amplitude` module.


expertsystem/reaction/solving.py
```diff
@@ -146,6 +146,41 @@ def extend(
                 else:
                     self.__violated_rules[key].update(rules2)
 
+    def get_initial_state(self) -> List[Particle]:
+        if len(self.solutions) == 0:
+            raise ValueError(
+                f"No solutions in {self.__class__.__name__} object"
+            )
```
spflueger, line 153: Thats already done in `__get_first_graph()`

expertsystem/reaction/solving.py
```diff
@@ -146,6 +146,41 @@ def extend(
...
+        graph = self.__get_first_graph()
+        return [
+            graph.edge_props[i][0] for i in graph.get_initial_state_edges()
+        ]
+
+    def get_final_state(self) -> List[Particle]:
+        if len(self.solutions) == 0:
+            raise ValueError(
+                f"No solutions in {self.__class__.__name__} object"
+            )
```
spflueger, line 163: Same

spflueger (commented):
The whole Result class still feels strange. I will approve the PR but we should think about the design in total. I don't like that Result is passed (with violated rules info) into the amp builders. Currently I'm thinking about slipping up the Result class. Then as a result a simple tuple of the solutions and the violated rules is returned. If a solver cannot execute all rules, then an exception could be raised instead of returning that information in the Result as well. I think that was the original design.... maybe that better then trying to combine everything in this Result class.

expertsystem/reaction/solving.py
redeboer: Ah thanks, weird mistake

redeboer:
> The whole Result class still feels strange

I agree. It's a bit related to #325 though, that I think it's good to return a 'navigable' and safe class, as in, you can probe its contents through methods and it is ensured that it only contains `StateTransitionGraph` instances with consistent initial and final state. But that's more narrowed down than a `Result` class (too general naming btw) that has the responsibility (?) of keeping track of which rules were violated.

redeboer:
Actually for #274, I convert the `Result` back into an exception, which carries all violated rules as `args` of the exception. So that's already a code smell I guess

spflueger (approved):

Issue #303, opened 2020-10-08
Define Rules which apply to nodes or edges individually


spflueger:
Some rules like the Gellmann Nishijima Rule are subject only to edge properties. The new spin validity rules fall under this category as well. Rules that only work with node properties or edge properties are needed. This simplifies and speeds up the code.

Also, these rules have to be optional in the sense that if no isospin is defined, then also the isospin validity check has to pass.

- [x] Create two new rule Protocols.
- [x] Change rules that only require edge or node quantum numbers to satisfy the new Protocols.
- [x] Adapt solving code to the new edge and node only rules.
- [x] ~~Replace the Gellmann Nishijima checking code in Particle to use the new rule instead.~~

Unfortunately with the solving interface using `Particle` introduces a cyclic dependency when trying to use the Gellmann Nishijima rule in the Particle initialization. This can be fixed after #369. Moving this task into a new issue.

spflueger:
With PR #331 this issue can be handled more easily now. The idea is to introduce two new `Protocols` that allow Rules with single arguments of type `Edge-` and `NodeQuantumNumbers`, and adapt the solving code to link edges and nodes individually to these rules.

PR #331, opened 2020-10-22
refactor!: improve Rule argument handling code

spflueger:
Extract the Rule argument handling code into a separate module. Also allow more flexible arguments and improve type annotation handling via inspect. Exchange the Rule interface with a `Protocol` (`ConservationRule`) and define a second type of Rule (`EdgeQNConservationRule`), which does not require any node quantum numbers.

This prepares things for issue #303

### DIFF
--- a/expertsystem/reaction/conservation_rules.py
+++ b/expertsystem/reaction/conservation_rules.py
@@ -8,10 +8,10 @@ There are four different types of rules:
 
 1. Rules that work on individual graph edges (WIP).
 2. Rules that work on individual graph nodes (WIP).
-3. Rules that work on the interaction level, which use ingoing edges, outgoing
-   edges as arguments. For example `.ChargeConservation`.
-4. Rules that work on the interaction level, which use ingoing edges, outgoing
-   edges and a interaction node as arguments. For example
+3. `EdgeQNConservationRule` that work on the interaction level, which use
+   ingoing edges, outgoing edges as arguments.  E.g.: `.ChargeConservation`.
+4. `ConservationRule` that work on the interaction level, which use ingoing
+   edges, outgoing edges and a interaction node as arguments. E.g:
    `.parity_conservation`.
 
 The arguments can be any type of quantum number. However a rule argument
--- a/expertsystem/reaction/conservation_rules.py
+++ b/expertsystem/reaction/conservation_rules.py
@@ -1,20 +1,55 @@
 """Collection of quantum number conservation rules for particle reactions.
 
 This module is the place where the 'expert' defines the rules that verify
-quantum numbers of the reaction. The module is therefore strongly typed (both
+quantum numbers of the reaction.
+
+A rule is a function that takes quantum numbers as input and outputs a boolean.
+There are four different types of rules:
+
+1. Rules that work on individual graph edges (WIP).
+2. Rules that work on individual graph nodes (WIP).
+3. Rules that work on the interaction level, which use ingoing edges, outgoing
+   edges as arguments. For example `.ChargeConservation`.
+4. Rules that work on the interaction level, which use ingoing edges, outgoing
+   edges and a interaction node as arguments. For example
+   `.parity_conservation`.
+
+The arguments can be any type of quantum number. However a rule argument
+resembling edges only accepts `~.quantum_numbers.EdgeQuantumNumbers`. Similarly
+arguments that resemble a node only accept
+`~.quantum_numbers.NodeQuantumNumbers`. The argument types do not have to be
+limited to a single quantum number, but can be a composite (see
+`.CParityEdgeInput`).
+
+.. warning::
+    Besides the rule logic itself, a rule also has the responsibility of
+    stating its run conditions. These run conditions **must** be stated by
+    the type annotations of its **__call__** method. The type annotations
+    therefore are not just there for *static* type checking: they also
+    carry more information about the rule that is extracted *dynamically*
+    by the `.reaction` module.
+
+Generally, the conditions can be separated into two categories:
+
+* variable conditions
+* toplogical conditions
+
+Currently, only variable conditions are being used. Topological conditions
+could be created in the form of `~typing.Tuple` instead of `~typing.List`.
+
+For additive quantum numbers, the decorator `additive_quantum_number_rule`
+can be used to automatically generate the appropriate behavior.
+
+
+The module is therefore strongly typed (both
 for the reader of the code and for type checking with :doc:`mypy
 <mypy:index>`). An example is `.HelicityParityEdgeInput`, which has been
-defined to provide type checks on `.ParityConservationHelicity`.
-
-See more information under `Rule`.
+defined to provide type checks on `.parity_conservation_helicity`.
 """
-
-# pylint: disable=abstract-method
-
 
 from copy import deepcopy
 from functools import reduce
-from typing import Any, Callable, List, Optional, Set, Tuple, Type, Union
+from typing import Any, Callable, List, Optional, Set, Tuple, Union
 
 import attr
 
@@ -22,6 +57,11 @@ from expertsystem.particle import Spin
 
 from .combinatorics import arange
 from .quantum_numbers import EdgeQuantumNumbers, NodeQuantumNumbers
+
+try:
+    from typing import Protocol
+except ImportError:
+    from typing_extensions import Protocol  # type: ignore
 
 
 def _is_boson(spin_magnitude: float) -> bool:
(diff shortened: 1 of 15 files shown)


expertsystem/reaction/conservation_rules.py
```diff
@@ -229,7 +237,7 @@ class CParityNodeInput:
     s_mag: NodeQuantumNumbers.s_magnitude = attr.ib()
 
 
-class CParityConservation(Rule):
+class CParityConservation:
```
redeboer, line 240: So here it's important to be aware that the developer is not warned  anymore (through `mypy` or other linters) within this module when a signature is incorrect. If I understand correctly, the `Protocol` is only enforced statically at the place where the rule is called, wereas with an ABC (`Rule`, previously), the signature is checked at the place of the definition.

expertsystem/reaction/conservation_rules.py
```diff
@@ -735,26 +746,25 @@ def __call__(
...
-    .. math::
-        |S_1 - S_2| \leq S \leq |S_1 + S_2|
-
-    and
-
-    .. math::
-        |L - S| \leq J \leq |L + S|
-    """
-
+class SpinMagnitudeConservation:
```
redeboer, line 749: Yes, that's indeed better naming

expertsystem/reaction/solving.py
```diff
@@ -83,14 +91,37 @@ class GraphSettings:
     node_settings: Dict[int, NodeSettings] = attr.ib(factory=dict)
 
 
+# a tuple of rules defines a group which together violates all possible
```
redeboer, line 94: Why a comment, not a docstring?

expertsystem/reaction/solving.py
```diff
@@ -83,14 +91,37 @@ class GraphSettings:
...
+# combinations that were processed
+def convert_to_names(
+    rules: Dict[int, Set[Tuple[Rule, ...]]]
+) -> Dict[int, Set[Tuple[str, ...]]]:
```
redeboer, line 98: Why convert to a `str`?

expertsystem/reaction/solving.py
```diff
@@ -746,16 +562,15 @@ def get_rules_by_priority(
             # and strip away the priorities again
             return [x[0] for x in sorted_list]
 
+        arg_handler = RuleArgumentHandler()
+
         for node_id in self.__graph.nodes:
-            # currently we only have rules related to graph nodes
```
redeboer, line 750: Thinking about this comment. Did it become outdated because of the new distinction between `EdgeInteractionRule` and `InteractionRule`?

expertsystem/reaction/conservation_rules.py
```diff
@@ -297,7 +305,7 @@ class GParityNodeInput:
     s_mag: NodeQuantumNumbers.s_magnitude = attr.ib()
 
 
-class GParityConservation(Rule):
+class GParityConservation:
```
redeboer, line 308: Wait, why do we still use classes now that there are no abstract base classes anymore? There was some problem with the signatures or so, right?

(Can also be handled in a follow-up issue)

tests/unit/reaction/conservation_rules/test_additive_qn.py
```diff
@@ -8,6 +8,4 @@
     [(([0], [-1, 1]), True), (([0], [1, 1]), False)],
 )
 def test_charge_conservation(graph_input, expected_value):
-    rule = ChargeConservation()
-
-    assert rule(*graph_input) == expected_value
+    assert ChargeConservation()(*graph_input) == expected_value
```
redeboer, line 11: See question on why the rules are still classes.

redeboer (commented):
I added some thoughts inline. Overall I think that `Protocol` (static duck typing [PEP 544](https://www.python.org/dev/peps/pep-0544)) is the way to go here. I'm still not 100% convinced though, because the typing takes place outside the module where the rules are defined. But it does seem that `Protocol` offers some more flexibility (I'm for instance also thinking about `kwargs`, but they are not needed at this stage).

Note: as discussed earlier, the strong emphasis on typing we introduced over the passed releases, seems to have caused quite some pain within the `conservation_rules` and `solving` modules. That's what I mean with flexibility, that I hope that `Protocol` brings back some of the duck typing behaviour that we used to have, but still with the possibility to type as much as possible.

expertsystem/reaction/conservation_rules.py
```diff
@@ -1,20 +1,54 @@
 """Collection of quantum number conservation rules for particle reactions.
 
 This module is the place where the 'expert' defines the rules that verify
-quantum numbers of the reaction. The module is therefore strongly typed (both
+quantum numbers of the reaction.
+
+A rule is a function that takes quantum numbers as input and outputs a boolean.
+There are three different types of rules:
+
+1. Rules that work on individual graph edges.
```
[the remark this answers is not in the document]
spflueger: yeah, i did not spend much effort on the doc strings yet as its still a bit experimental PR

expertsystem/reaction/solving.py
spflueger: This is related to the issue of uniquely identifying a rule. The user is only interested why a problem does not have any solutions, so for example because of c parity violation. A string is enough for this. Currently the rule objects are used themselves, and define a human readable str (repr) method as well as a hash function. This is not very transparent.

expertsystem/reaction/solving.py
spflueger: Partly, especially the new upcoming edge rules will make it different

redeboer: Upcoming edge rules?

expertsystem/reaction/conservation_rules.py
spflueger: This is related to the fact that I was not able to get the Protocols working for regular functions.

tests/unit/reaction/conservation_rules/test_additive_qn.py
spflueger: The rules created with the additive quantum number decorator have the additional difficulty to achieve the pickling support. With classes this much easier to sidestep

expertsystem/reaction/conservation_rules.py
spflueger: Since I now know that the full signature of the protocol functions is important, also the argument names, I can verify that it also works with regular functions. So this can be changed

tests/unit/reaction/conservation_rules/test_additive_qn.py
redeboer: Ah yes, the pickling was the problem

spflueger:
> I added some thoughts inline. Overall I think that `Protocol` (static duck typing [PEP 544](https://www.python.org/dev/peps/pep-0544)) is the way to go here. I'm still not 100% convinced though, because the typing takes place outside the module where the rules are defined. But it does seem that `Protocol` offers some more flexibility (I'm for instance also thinking about `kwargs`, but they are not needed at this stage).
> 
> Note: as discussed earlier, the strong emphasis on typing we introduced over the passed releases, seems to have caused quite some pain within the `conservation_rules` and `solving` modules. That's what I mean with flexibility, that I hope that `Protocol` brings back some of the duck typing behaviour that we used to have, but still with the possibility to type as much as possible.

Thanks for the review. I'm mainly just unhappy with the rather imprecise rule interface/protocols. A precise definition of a rule in terms of an interface is needed to run the solving process. However our need is also the rather flexible definition of rules. Rule arguments can vary from quantum numbers of a single edge, to quantum numbers from all edge going in and out of a node, as well as quantum numbers from the node itself. Also it would be preferable if a rule can chose the names for arguments and they are not predetermined by the interface.
So the concept of an interface is not very suitable here. However I currently do not have any good concept to handle the solving without an interface of a rule. The hard part is to feel the quantum numbers of a state transition graph to a rule. So knowing which quantum numbers a rule wants or needs in order to execute, and where to find those in the graph

redeboer (approved):
Okay, GJ!
The only thing I find a bit counter-intuitive is the naming and in that extend the type of objects. This becomes particularly clear here:

https://github.com/ComPWA/expertsystem/blob/754666d3721d8ddfb461db398cce0fc7183c526f/expertsystem/reaction/_default_settings.py#L50-L73

Of course, this just reflects the fact that a switch from `ABC` to `Protocol` allows for duck typing, so I guess it's a matter of getting used to.

spflueger:
> Okay, GJ!
> The only thing I find a bit counter-intuitive is the naming and in that extend the type of objects. This becomes particularly clear here:
> 
> https://github.com/ComPWA/expertsystem/blob/754666d3721d8ddfb461db398cce0fc7183c526f/expertsystem/reaction/_default_settings.py#L50-L73
> 
> Of course, this just reflects the fact that a switch from `ABC` to `Protocol` allows for duck typing, so I guess it's a matter of getting used to.

Yes, I would also say a matter of getting used to. Lets see how this works for us, we can always still modify it.

PR #336, opened 2020-10-24
feat: implement Result.get_particle_graphs

redeboer:
Extract a `List[StateTransitionGraph[Particle]]` (without spin projections) from the `Result` class. This is useful when visualising solutions, e.g.:
image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,7 @@
 {
+  "[ipynb]": {
+    "editor.formatOnSave": false
+  },
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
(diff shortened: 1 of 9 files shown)


redeboer:
I'm thinking about adding an option to merge the graphs even further. So in this case, that would result in only one graph with the label `"f(0)(500)\n...\nf(0)(1710)"`.

redeboer:
> I'm thinking about adding an option to merge the graphs even further. So in this case, that would result in only one graph with the label `"f(0)(500)\n...\nf(0)(1710)"`.

I have some solution for this, but I'll handle this in a follow-up PR

redeboer:
This PR is a bit related to comments like https://github.com/ComPWA/expertsystem/pull/328#pullrequestreview-512083599, but while the discussion on the `Result` class is not settled, it's best to implement functionality as this there.

PR #340, opened 2020-10-28
fix: sort ParticleColl. getitem candidates by mass

redeboer:
Small enhancement I found to be useful when defining the initial and final states and making a typo. Results in the following https://github.com/ComPWA/expertsystem/blob/64d4ec18028063ccc1f2226925e545a5e65afd22/tests/unit/test_particles.py#L342-L350


sebastianJaeger (approved):
Looks good :)

Issue #169, opened 2020-08-06
Make it easier to set default maximum values


redeboer:
At the moment, all default maximum values are defined under `ui._default_settings`. This is hard to modify as a user (strictly speaking not allowed according to Python conventions). An interface should be created that makes it easier to set these maximum values.

Note that these values are important for performance, especially when a large number of resonances is to be considered.

PR #368, opened 2020-11-04
feat: support pure edge and node rules in solving

### LINKED ISSUE #303: Define Rules which apply to nodes or edges individually

spflueger:
Closes #303 

Note that the solving code is not very DRY. But to keep the PR reasonable and make progress faster I chose to have some repetition. This can be improved in another PR

### DIFF
--- a/src/expertsystem/reaction/__init__.py
+++ b/src/expertsystem/reaction/__init__.py
@@ -396,7 +396,6 @@ class StateTransitionManager:  # pylint: disable=too-many-instance-attributes
                 f"after qn solving: {len(result.solutions)}",
             )
 
-        # merge strengths
         final_result = Result()
         for temp_result in results.values():
             final_result.extend(temp_result)
--- a/src/expertsystem/reaction/__init__.py
+++ b/src/expertsystem/reaction/__init__.py
@@ -278,7 +278,17 @@ class StateTransitionManager:  # pylint: disable=too-many-instance-attributes
         # pylint: disable=too-many-locals
         final_state_edges = graph.get_final_state_edges()
         initial_state_edges = graph.get_initial_state_edges()
-        graph_settings: List[GraphSettings] = [GraphSettings({}, {})]
+        graph_settings: List[GraphSettings] = [
+            GraphSettings(
+                edge_settings={
+                    edge_id: self.interaction_type_settings[
+                        InteractionTypes.Weak
+                    ][0]
+                    for edge_id in graph.edges
+                },
+                node_settings={},
+            )
+        ]
 
         for node_id in graph.nodes:
             interaction_types: List[InteractionTypes] = []
@@ -323,9 +333,6 @@ class StateTransitionManager:  # pylint: disable=too-many-instance-attributes
             for temp_setting in temp_graph_settings:
                 for int_type in interaction_types:
                     updated_setting = deepcopy(temp_setting)
-                    updated_setting.edge_settings[node_id] = deepcopy(
-                        self.interaction_type_settings[int_type][0]
-                    )
                     updated_setting.node_settings[node_id] = deepcopy(
                         self.interaction_type_settings[int_type][1]
                     )
@@ -405,8 +412,10 @@ class StateTransitionManager:  # pylint: disable=too-many-instance-attributes
             match_external_edges(final_solutions)
         return Result(
             final_solutions,
-            final_result.not_executed_rules,
-            final_result.violated_rules,
+            final_result.not_executed_node_rules,
+            final_result.violated_node_rules,
+            final_result.not_executed_edge_rules,
+            final_result.violated_edge_rules,
             formalism_type=self.formalism_type,
         )
 
@@ -596,14 +605,17 @@ def check(
         topology_building=topology_building,
     )
 
-    if any(map(len, results.violated_rules.values())):
+    if any(map(len, results.violated_node_rules.values())) or any(
+        map(len, results.violated_edge_rules.values())
+    ):
         if isinstance(initial_state, list):
             initial_state = " ".join(initial_state)
         if isinstance(final_state, list):
             final_state = " ".join(final_state)
         raise ValueError(
             f'Reaction "{initial_state} -> {final_state}" violates',
-            results.violated_rules,
+            results.violated_node_rules,
+            results.violated_edge_rules,
             "when using interaction types",
             allowed_interactions,
         )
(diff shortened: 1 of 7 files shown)


src/expertsystem/reaction/__init__.py
```diff
@@ -278,7 +278,17 @@ def _determine_graph_settings(
         # pylint: disable=too-many-locals
         final_state_edges = graph.get_final_state_edges()
         initial_state_edges = graph.get_initial_state_edges()
-        graph_settings: List[GraphSettings] = [GraphSettings({}, {})]
+        graph_settings: List[GraphSettings] = [
+            GraphSettings(
+                edge_settings={
+                    edge_id: self.interaction_type_settings[
+                        InteractionTypes.Weak
```
redeboer, line 285: Why weak by default?

src/expertsystem/reaction/__init__.py
```diff
@@ -405,8 +411,10 @@ def find_solutions(
             match_external_edges(final_solutions)
         return Result(
             final_solutions,
-            final_result.not_executed_rules,
-            final_result.violated_rules,
+            final_result.not_executed_node_rules,
```
redeboer, line 414: We indeed have to split up `Result`. That (as you suggested), `find_solutions` returns a `StateTransition` (#325) and some `ViolatedRules` that bundles the edge and node rules (possibly also the `not_executed_rules`, not sure).

src/expertsystem/reaction/__init__.py
```diff
@@ -596,14 +604,17 @@ def check(
         topology_building=topology_building,
     )
 
-    if any(map(len, results.violated_rules.values())):
+    if any(map(len, results.violated_node_rules.values())) or any(
```
redeboer, line 607: Hmm maybe
```python
violated_rules = (
    results.violated_node_rules.values()
    + results.violated_edge_rules.values()
)
if any(map(len, violated_rules)):
    ...
```
Ah but there we go again, this `Result` class is just fishy...

src/expertsystem/reaction/_default_settings.py
```diff
@@ -105,7 +114,35 @@ def create_default_interaction_settings(
     E.g.: strong, em and weak interaction.
     """
     interaction_type_settings = {}
-    formalism_edge_settings = EdgeSettings()
+    formalism_edge_settings = EdgeSettings(
```
redeboer, line 117: #169

src/expertsystem/reaction/conservation_rules.py
```diff
@@ -4,13 +4,12 @@
...
 
 A rule is a function that takes quantum numbers as input and outputs a boolean.
-There are four different types of rules:
+There are three different types of rules:
 
-1. Rules that work on individual graph edges (WIP).
-2. Rules that work on individual graph nodes (WIP).
-3. `EdgeQNConservationRule` that work on the interaction level, which use
+1. `GraphElementRule` that work on individual graph edges or nodes.
+2. `EdgeQNConservationRule` that work on the interaction level, which use
```
redeboer, line 10: Perhaps for the 'symmetry' between nodes and edges we need to think of naming here: `EdgeConservationRule` and `NodeConservationRule`. But not sure if I understand the intention correctly.

src/expertsystem/reaction/conservation_rules.py
```diff
@@ -422,8 +424,14 @@ def _check_particles_identical(
     return True
 
 
+@attr.s(frozen=True)
```
redeboer, line 427: I guess in preparation of #369?

redeboer (approved):
Good job!
I find it hard to judge the changes to the `solving` module, but I like additions like the `Scoresheet` for instance. My only worry is that the module increased by 300 lines, but if I understand correctly that is due to issues like #369.

src/expertsystem/reaction/__init__.py
spflueger: I tried that as well, but you cant add generators, so I would have to wrap each part into a `list()`.

src/expertsystem/reaction/__init__.py
spflueger: That's indeed quite misleading... The thing is that there are interaction rules, which belong to an interaction and must involve edges going into AND out of a node. The other part of rules are just consistency checkers, that the information attached to a single edge or node are valid from a physics definition, so independent of interactions. Like checking that the spin magnitude and projection fit together or the Gellmann Nishijima rule.

src/expertsystem/reaction/__init__.py
spflueger: We need an issue for this where we can discuss the various possibilities and problems related to this.

src/expertsystem/reaction/conservation_rules.py
spflueger: See https://github.com/ComPWA/expertsystem/pull/368#discussion_r517947995

src/expertsystem/reaction/__init__.py
redeboer: I think #325 suffices, only need to reformulate it a bit

src/expertsystem/reaction/conservation_rules.py
spflueger: Yes. I my view, the reaction/solving part has to purely work on individual quantum numbers and never create any composite quantum number data structures with logic inside them. The logic should be handled by the rules.
Note that this is a rather subtle but yet largely effecting change. Once spin validation rule verifies the correctness of a spin, it does not need to be checked over and over again. That previously happened, since the regular Spin data structure was used. Alternatively one could also just pass the individual quantum numbers around, but its more convenient to work with this composite Spin data structure.

PR #373, opened 2020-11-06
feat: add reaction violation checker

### LINKED ISSUE #281: Implement check for all violations in a reaction
A new function that can check for all violated rules is needed. I propose the form 
```python3 
check_violations(initial_state, final_state, interaction_type: Optional[InteractionTypes]) -> List[str]
```

Before the `FullPropagator` also had this responsibility of finding all violations. This functionality has to be re-implemented here. Concretely an additional managing layer on top of the csp solver is needed. There are two options (I favor option 1):

1. A new approach would be to always use all rules and individually check each rule which graph they violate. Then by looking at the results of each rule the violation can be accurately determined. If the user requested a specific interaction, the violated rules can be filtered for that specific interaction.
2. Previously it was iteration on the temporary result until there are no more violations. This iteration is required due to early stopping of the solvers if there are violations. The following already implements this (it just has to be adapted a bit)

```python3
        def __determine_all_violations(
            self, current_result: Result, graph_settings: GraphSettings
        ) -> Result:
            # Unfortunately the csp constraint module will jump out of the
            # solving process one a rule violates all possibilities.
            # In order to find all violations a recursing with csp is needed.
            # A simple validation will not suffice, because some node quantum
            # numbers may not be set by the user. So they have to be generated
            # by the system.
            # Stop rerun once the set of not_executed_rules does not change
            # or is empty. Then just finish up with validation
            # solver = CSPSolver(self.__allowed_intermediate_particles)
            previous_result = Result([])
            violated_rules = current_result.violated_rules
            while (
                previous_result.not_executed_rules
                != current_result.not_executed_rules
                and current_result.not_executed_rules
            ):
                previous_result = current_result
                for (
                    node_id,
                    node_setting,
                ) in graph_settings.node_settings.items():
                    node_setting.conservation_rules = (
                        current_result.not_executed_rules[node_id]
                    )
                current_result = self.__run_csp(self.__graph, graph_settings)
                for node_id, rule_set in current_result.violated_rules.items():
                    violated_rules[node_id].update(rule_set)

            validated_result = validate_fully_initialized_graph(
                self.__graph, current_result.not_executed_rules
            )
            for (
                node_id,
                rule_set,
            ) in validated_result.violated_rules.items():
                violated_rules[node_id].update(rule_set)

            return Result(
                [], validated_result.not_executed_rules, violated_rules
            )
```

spflueger:
Closes #281 

Note that it currently cannot guarantee to find all violations on a general level, as this may require to run the CSPSolver with intermediate states. With the current two stage processing workflow this extremely time consuming.


spflueger:
Note that I still have to improve 
- [] the docstring
- [] add the rule group violation feature (a group of rules violate things as a collection)

src/expertsystem/reaction/__init__.py
```diff
@@ -429,6 +453,184 @@ def _solve(
...
+    ) -> Set[str]:
+        node_id = list(graph.nodes)[0]
+        return validate_fully_initialized_graph(
+            graph,
+            rules_per_node={node_id: node_rules},
+            rules_per_edge={},
+        ).violated_node_rules[node_id]
+
+    # Step 1: create n-body topology
+    # Using and n-body topology is enough, to determine the violations reliably
```
redeboer, line 491: Using an~d~

src/expertsystem/reaction/__init__.py
```diff
@@ -429,6 +453,184 @@ def _solve(
...
 
 
+def check_reaction_violations(
+    initial_state: Union[StateDefinition, Sequence[StateDefinition]],
+    final_state: Sequence[StateDefinition],
+) -> Set[Tuple[str, ...]]:
+    """Determine violated interaction rules for a given particle reaction.
+
+    .. warning:: This function does only guarantees to find P, C and G parity
+      violations, if its a two body decay. If all initial and final states have
```
redeboer, line 463: it's

src/expertsystem/reaction/__init__.py
```diff
@@ -429,6 +453,184 @@ def _solve(
...
+        len(initial_state), len(final_state)
+    )[0]
+
+    # Step 2: initialize topologies with initial and final state
+    initialized_graphs = initialize_graph(
+        topology=topology,
+        particles=load_default_particles(),
+        initial_state=initial_state,
+        final_state=final_state,
+    )
```
redeboer, line 513: Perhaps like in line 505 also take `[0]` here?

src/expertsystem/reaction/__init__.py
```diff
@@ -429,6 +453,184 @@ def _solve(
...
+        rules_per_node={},
+        rules_per_edge={
+            edge_id: pure_edge_rules
+            for edge_id in initialized_graphs[0].get_initial_state_edges()
+            + initialized_graphs[0].get_final_state_edges()
+        },
+    )
+
+    if edge_check_result.violated_edge_rules:
+        raise ValueError(
```
redeboer, line 533: Why raise here?

src/expertsystem/reaction/__init__.py
```diff
@@ -429,6 +453,184 @@ def _solve(
...
+            f"Some edges violate"
+            f" {edge_check_result.violated_edge_rules.values()}"
+        )
+
+    # Those rules give the same results, independent on the node and spin props.
+    # Note they are also independent of the topology and hence their results
+    # are always correct.
+    edge_qn_conservation_rules: Set[Rule] = {
```
redeboer, line 541: Hm not very DRY, but once again related to #169

src/expertsystem/reaction/__init__.py
```diff
@@ -429,6 +453,184 @@ def _solve(
...
+        )
+    }
+
+    # Step 5: Create combinations of graphs for magnitudes of S and L, but only
+    # if it is a two body reaction
+    ls_combinations = [
+        InteractionProperties(l_magnitude=l_mag, s_magnitude=s_mag)
+        for l_mag, s_mag in product([0, 1], [0, 0.5, 1, 1.5, 2])
+    ]
+    node_id = list(initialized_graphs[0].nodes)[0]
```
redeboer, line 568: Alternative without conversion:
```python
next(iter(initialized_graphs[0].nodes))
```

redeboer: Also see remark at 513

src/expertsystem/reaction/__init__.py
```diff
@@ -429,6 +453,184 @@ def _solve(
...
+    graphs = []
+    for ls_combi in ls_combinations:
+        for graph in initialized_graphs:
+            new_graph = copy(graph)
+            new_graph.node_props = {node_id: ls_combi}
+            graphs.append(new_graph)
+
+    # Step 6: Then verify each graph with the interaction rules. Process each
+    # rule separately and keep track of the on which graph each rule passes or
```
redeboer, line 577: ~the on~

tests/channels/test_nbody_reactions.py
```diff
@@ -27,164 +25,138 @@ def reduce_violated_rules(
...
-        #     ],
-        # ),
-        ((["Delta(1232)+"], ["p", "pi0"]), []),
-        ((["nu(e)~", "p"], ["n", "e+"]), []),
-        # (  # https://github.com/ComPWA/expertsystem/issues/281
-        #     (["e-", "p"], ["nu(e)", "pi0"]),
-        #     ["BaryonNumberConservation", "SpinConservation"],
-        # ),
-        ((["p", "p"], ["Sigma+", "n", "K~0", "pi+", "pi0"]), []),
+            {"ChargeConservation", "isospin_conservation"},
```
redeboer, line 28: Though not immediately relevant for this PR (the conversion from `ConservationRule` class to `str` was introduced earlier on), I now wonder if the strings should be made more concise/intuitive, like `violated_rules == {"isospin", "charge", "c_parity"}`. The naming distinction between function vs callable class is irrelevant for the output of this facade.

tests/channels/test_nbody_reactions.py
```diff
@@ -27,164 +25,138 @@ def reduce_violated_rules(
...
 
-    stm.set_allowed_interaction_types([InteractionTypes.Strong])
-
-    graph_interaction_settings = stm.prepare_graphs()
-    result = stm.find_solutions(graph_interaction_settings)
-
-    reduced_violations = reduce_violated_rules(result.violated_node_rules)
-    assert set(reduced_violations) == set(expected)
+    reduced_violations = reduce_violated_rules(violations)
+    print("not allowed! violates: " + str(reduced_violations))
```
redeboer, line 161: I would just remove all print statements in tests (and replace them with `assert`s where possible). Same with line 153

redeboer (commented):
Here are some thoughts already.
Some other problems I have (but no solution for it):
- It's not so clear to me what the nested `set`-of-`tuple`s is useful for.
- The function is non-configurable: rules are hard-coded into the function and don't make use of the default settings.
- I'm trying to see whether it's possible to split out the different steps (having a long function that needs to be explained in comments comments), but it's hard to see how.

tests/channels/test_nbody_reactions.py
spflueger: Can of course be done, but some conversion code or `__str__` definitions

src/expertsystem/reaction/__init__.py
spflueger: That's just a safety feature, that should never really happen. These edge rules actually would not needed to be checked, as they simply verify a particle has consistent quantum numbers. And since the particles are extracted from the particle collection and that separately guarantees that particles are clean. I just put that check in as a extra safety net. We could also think about removing the pure edge rules here completely

tests/channels/test_nbody_reactions.py
spflueger: I guess reliably only the latter works

src/expertsystem/reaction/__init__.py
spflueger: But due to spin projections multiple graphs are created here and all of them are needed at the end.

src/expertsystem/reaction/__init__.py
spflueger: That's true. I did not have a good and small solutions while coding this. Hence I solved it like this for now. What I can say is that it would be convenient for this check to retrieve rules as bundles, based on their type (edge rule, interaction rule etc.). Also the set of rules is a customized (it not just all rules split up by type).

src/expertsystem/reaction/__init__.py
spflueger: 44e8ea04bba93a7a35021d254c48a9afa1d045b3

src/expertsystem/reaction/__init__.py
redeboer: Ah I see now. At first I only saw calls to `[0]`

src/expertsystem/reaction/__init__.py
spflueger: 44e8ea04bba93a7a35021d254c48a9afa1d045b3

src/expertsystem/reaction/__init__.py
spflueger: 44e8ea04bba93a7a35021d254c48a9afa1d045b3

src/expertsystem/reaction/__init__.py
spflueger: 44e8ea04bba93a7a35021d254c48a9afa1d045b3

tests/channels/test_nbody_reactions.py
spflueger: 44e8ea04bba93a7a35021d254c48a9afa1d045b3

tests/channels/test_nbody_reactions.py
redeboer: We can leave it like this, it would just be a service to the user, but the current output strings speak for themselves

src/expertsystem/reaction/__init__.py
redeboer: Ok I misunderstood the logics there then

spflueger:
So the PR would be finished from my side. I tried to make the code a bit more self explanatory and get rid of the comments.

redeboer (approved):
GJ

redeboer (changes requested):
So one thing I just realised: there's this weird `reaction.check` function here
https://github.com/ComPWA/expertsystem/blob/44e8ea04bba93a7a35021d254c48a9afa1d045b3/src/expertsystem/reaction/__init__.py#L770-L813

I say weird because I implemented it :grimacing: and know its return values / exceptions raised are counterintuitive. It can be removed in favour of the new violation checker.

However, the alias at the the top module also needs to be replaced somehow
https://github.com/ComPWA/expertsystem/blob/eceeb4bbf8f90954c63040a49493e8a13963044f/src/expertsystem/__init__.py#L54-L55

and that means the interactive page (notebook) needs to be rewritten a bit.

spflueger:
Ok changed

redeboer (approved):
Removed the `reaction.check` facade just now, but otherwise approved :+1:

PR #374, opened 2020-11-08
ci: define GitHub workflow for epic branches

redeboer:
This is an extension of #360. The idea is to have the latest 'stable' state of the framework in the `master` branch and isolate bigger reworks to an 'epic' branch of the form `epic/*`. The CI for this branch is under 1 minute and does not require review, but still allows for discussion.


spflueger:
I guess we have to try this branching model and see if it works like that

spflueger (approved):

Issue #377, opened 2020-11-10
Fix CD and install issues


redeboer:
There are a few bugs with the deployment:
- [x] For some reason, the CD job does not run anymore upon creating a release
- [x] The notebooks don't work anymore on Google Colab. When running `pip install expertsystem[doc]`, one gets a PEP517 error
  ```log
  Collecting expertsystem[doc]
    Downloading https://files.pythonhosted.org/packages/39/a0/071521c5b2dc67aca80ad925634b29d77e902a54ffe8a909d25fe7899c89/expertsystem-0.6.2.tar.gz (125kB)
      |████████████████████████████████| 133kB 2.7MB/s 
    Installing build dependencies ... done
    Getting requirements to build wheel ... done
      Preparing wheel metadata ... error
  ERROR: Command errored out with exit status 1: /usr/bin/python3 /usr/local/lib/python3.6/dist-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmpmt7g_nci Check the logs for full command output.
  ```
  I think this can be solved by building a wheel

redeboer:
Update: installing the `expertsystem` on Google Colab is again possible.

PR #376, opened 2020-11-10
fix: clebsch gordan helicity to canonical rule

spflueger:

PR #378, opened 2020-11-10
fix: build wheel and fix deploy job

redeboer:
Here's a first attempt at (partially) fixing #377. I'm not fully sure whether it works, but we want to have the #376 released for `tensorwaves`.

To be clear, trying to fix two things here:
- [x] Let the [CD GitHub workflow](https://github.com/ComPWA/expertsystem/actions?query=workflow%3ACD) trigger again
- [x] Make the package on PyPI PEP517 compatible (see problem with Colab in #377).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -6,6 +6,9 @@ on:
   create:
     tags:
       - "[0-9]+.[0-9]+.[0-9]+*"
+  release:
+    types:
+      - created
 
 jobs:
   test:
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,4 +1,4 @@
-# cspell:ignore noreply pypa sdist
+# cspell:ignore bdist noreply pypa sdist
 
 name: CD
 
@@ -114,7 +114,7 @@ jobs:
         with:
           python-version: 3.8
       - name: Create distribution
-        run: python setup.py sdist
+        run: python setup.py sdist bdist_wheel
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@master
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -114,7 +117,7 @@ jobs:
         with:
           python-version: 3.8
       - name: Create distribution
-        run: python setup.py sdist
+        run: python setup.py sdist bdist_wheel
```
spflueger, line 120: Why do we need a binary distribution?

spflueger: https://pythonwheels.com/ ok i read a bit here

.github/workflows/cd.yml
```diff
@@ -1,11 +1,14 @@
...
 
 name: CD
 
 on:
   create:
     tags:
       - "[0-9]+.[0-9]+.[0-9]+*"
+  release:
+    types:
+      - created
```
spflueger, line 11: why do we still need the `create: tags: ...` part?

redeboer: Actually I prefer the tags trigger, as that allows filtering on semantic versions. But tbh I don't understand why the existing tag trigger does not work anymore. That's why I added a release trigger as well.

.github/workflows/cd.yml
redeboer: Yeah and I suspect this is why it's not possible to run
```ipython
!pip install expertsystem==0.6.2
```
on Google Colab

.github/workflows/cd.yml
spflueger: Hmm ok

spflueger (approved):

PR #381, opened 2020-11-11
test: add unit tests for all conservation rules

### LINKED ISSUE #268: Write more unit tests for conservation_rules module
Currently not covered by the unit tests are:

- [x] `ParityConservation`
- [x] `ParityConservationHelicity`
- [x] `IdenticalParticleSymmetrization`
- [x] `ClebschGordanCheckHelicityToCanonical`
- [x] `HelicityConservation`
- [x] `GellMannNishijimaRule`

spflueger:
Closes #268 

Note that this may in the future avoid issues like #376

### DIFF
--- a/src/expertsystem/reaction/conservation_rules.py
+++ b/src/expertsystem/reaction/conservation_rules.py
@@ -185,7 +185,7 @@ def parity_conservation(
 
 @attr.s(frozen=True)
 class HelicityParityEdgeInput:
-    parity: Optional[EdgeQuantumNumbers.parity] = attr.ib()
+    parity: EdgeQuantumNumbers.parity = attr.ib()
     spin_mag: EdgeQuantumNumbers.spin_magnitude = attr.ib()
     spin_proj: EdgeQuantumNumbers.spin_projection = attr.ib()
 
@@ -202,8 +202,10 @@ def parity_conservation_helicity(
     .. math:: A_{-\lambda_1-\lambda_2} = P_1 P_2 P_3 (-1)^{S_2+S_3-S_1}
         A_{\lambda_1\lambda_2}
 
-    Notice that only the special case :math:`\lambda_1=\lambda_2=0` may
-    return False.
+    .. math:: \mathrm{parity\,prefactor} = P_1 P_2 P_3 (-1)^{S_2+S_3-S_1}
+
+    .. note:: Only the special case :math:`\lambda_1=\lambda_2=0` may
+      return False independent on the parity prefactor.
     """
     if len(ingoing_edge_qns) == 1 and len(outgoing_edge_qns) == 2:
         out_spins = [x.spin_mag for x in outgoing_edge_qns]
@@ -217,10 +219,11 @@ def parity_conservation_helicity(
             sum(out_spins) - ingoing_edge_qns[0].spin_mag
         )
 
-        daughter_hel = [0 for x in outgoing_edge_qns if x.spin_proj == 0.0]
-        if len(daughter_hel) == 2:
-            if prefactor == -1:
-                return False
+        if (
+            all([x.spin_proj == 0.0 for x in outgoing_edge_qns])
+            and prefactor == -1
+        ):
+            return False
 
         return prefactor == parity_prefactor
     return True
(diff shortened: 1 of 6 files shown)


src/expertsystem/reaction/conservation_rules.py
```diff
@@ -382,21 +385,20 @@ def check_g_parity_isobar(
...
     pid: EdgeQuantumNumbers.pid = attr.ib()
 
 
 def identical_particle_symmetrization(
-    ingoing_edge_qns: List[IdenticalParticleSymmetryEdgeInput],
-    outgoing_edge_qns: List[IdenticalParticleSymmetryEdgeInput],
+    ingoing_parities: List[EdgeQuantumNumbers.parity],
+    outgoing_edge_qns: List[IdenticalParticleSymmetryOutEdgeInput],
 ) -> bool:
     """Implementation of particle symmetrization."""
```
redeboer, line 398: What exactly is particle symmetrization?

redeboer (approved):
Looks good!

Some general thoughts:
- I wonder if in the long term it makes sense to write some notebooks on how to use conservation rules. Ideally, these should work standalone, right?
- All the separate test files under
  `tests/unit/reaction/conservation_rules`
  could ([should](https://pwa.readthedocs.io/projects/expertsystem/en/0.6.3/develop.html#testing)) be merged under
  `tests/unit/reaction/test_conservation_rules`

spflueger:
> * I wonder if in the long term it makes sense to write some notebooks on how to use conservation rules. Ideally, these should work standalone, right?

I don't think anyone will use those rules separately. If this will be the case for some reason. Then we can add some docs for that. Atm I do not see the need for that.

> * All the separate test files under
>   `tests/unit/reaction/conservation_rules`
>   could ([should](https://pwa.readthedocs.io/projects/expertsystem/en/0.6.3/develop.html#testing)) be merged under
>   `tests/unit/reaction/test_conservation_rules`

You mean to avoid that extra package layer and get a more fine grained test output log?

src/expertsystem/reaction/conservation_rules.py
spflueger: I can write a bit more detailed description

redeboer:
> You mean to avoid that extra package layer and get a more fine grained test output log?

In the 'develop page', I wrote a note that the structure of unit test should reflect that of the components they test. In this case, the module `conservation_rules` should get a test called `test_conservation_rules` and its functions (say, `identical_particle_symmetrization`), should be named accordingly (`test_identical_particle_symmetrization`). But it's a debatable convention and it's not enforced through style checks or so.

spflueger:
> > You mean to avoid that extra package layer and get a more fine grained test output log?
> 
> In the 'develop page', I wrote a note that the structure of unit test should reflect that of the components they test. In this case, the module `conservation_rules` should get a test called `test_conservation_rules` and its functions (say, `identical_particle_symmetrization`), should be named accordingly (`test_identical_particle_symmetrization`). But it's a debatable convention and it's not enforced through style checks or so.

I can see some benefits in that. For example, all of the single function test could be put in the `__init__.py`. That would make the structure nicer. However for some rules, I have more than one test. Each of those tests bundle specific cases together, for example separating fermions and bosons cases. Bundling that altogether again is useful. We could use a class, or just use a module like now. A module seems better in that case.

redeboer:
https://github.com/ComPWA/expertsystem/pull/381#issuecomment-725351302

Earlier on, I tried putting unit tests in the `src` folder, right along the modules that they test, so that there is some sort of enforcement of https://github.com/ComPWA/expertsystem/pull/381#issuecomment-725344704. But this introduces problems like having to filter the tests out when installing the package as well as losing the ability to apply different style checks to the `test_` files, see e.g.
https://github.com/ComPWA/expertsystem/blob/2ef4f442aa62cd18fbf6434acabacd83576048fd/setup.cfg#L123-L126

I would leave it like this, it's more of a thought, but I currently don't see a way how to enforce it automatically.

spflueger:
> For example, all of the single function test could be put in the `__init__.py`.

I just tried this, and it does not work. So I would prefer the current setup.

src/expertsystem/reaction/conservation_rules.py
spflueger: 6e9f296910162ffcb49c8ec753e5c0ed6f5d0398

PR #394, opened 2020-11-25
build!: define dev deps through requirements.txt

redeboer:
Addresses https://github.com/ComPWA/PWA-pages/issues/54

New dev installation procedure is described here
https://pwa--394.org.readthedocs.build/projects/expertsystem/en/394/install.html


spflueger (approved):
Looks good. We could think about running the a CI job, which sets up its environment through `pip install .`. That way we get constant feedback on how our compatibility is.

redeboer:
> We could think about running a CI job, which sets up its environment through pip install . (+ test.in requirements). That way we get constant feedback on how our compatibility is.

Sounds good, best put into the CD job, which is run upon releases (and merging into the `stable` branch).

Tricky part here would be that the `test.in` currently refer to `requirements.txt`, which is a pinned version of the package dependencies. So `test.in` would have to be split into something like:

```shell
# test-unpinned.in
pydot
pytest
pytest-cov
pytest-profiling
pytest-xdist
```

and 

```shell
# test-pinned.in
-r ../requirements.txt
-r test.in
```

Then the unpinned test job would have to install through something like

```bash
pip install . -r .reqs/test.in
```

Side note: at this stage we may again have to consider providing a `[test]` extras section, so you can still do `pip install .[test]` and run the framework any time. Important remains thought that we keep one source of truth for the contstraints, and in that case it would have to be the extras section under `setup.cfg`, not some `test.in` file.

spflueger (approved):
GJ

Issue #383, opened 2020-11-13
Remove io module


redeboer:
This sub-module is not DRY and therefore hard to maintain

redeboer:
Reopened due to https://github.com/ComPWA/expertsystem/pull/519#discussion_r594332768

spflueger:
**io module:** io logic is grouped in one place. Individual io logic can still be somewhere else, its just a ui facade.
**pros:**
- potentially easier to comprehend (this is highly subjective)
- keeping io import level high (just import expertsystem.io)

**cons:**
- extra UI code

**w/o io module:** io interface is directly in the module.
**pros:**
- potentially easier to comprehend (this is highly subjective)
- no extra code needed
- io functionality comes with import of the module itself

**cons:**
- io import level is mixed (but is grouped with the module, so usually no extra import is needed)

PR #395, opened 2020-11-26
fix: correctly dump Breit-Wigner

redeboer:
There was a bug when dumpin an `AmplitudeModel` containing a `RelativisticBreitWigner` to YAML or XML. This PR fixes it.

Note: better solution is through #383, but this is more of a hot-fix that is required for TensorWaves, see e.g.
https://github.com/ComPWA/tensorwaves/pull/179/checks?check_run_id=1454030793

### DIFF
--- a/tests/unit/amplitude/expected_recipe.yml
+++ b/tests/unit/amplitude/expected_recipe.yml
@@ -28,10 +28,26 @@ Parameters:
   - Name: &par5 Phase_J/psi(1S)_to_f(0)(1500)_0+gamma_1;f(0)(1500)_to_pi0_0+pi0_0;
     Type: Phase
     Value: 0.0
-  - Name: MesonRadius_J/psi(1S)
+  - Name: &par6 MesonRadius_J/psi(1S)
     Type: MesonRadius
     Value: 1.0
     Fix: true
+  - Name: &par7 Position_f(0)(980)
+    Value: 0.99
+  - Name: &par8 Width_f(0)(980)
+    Value: 0.06
+  - Name: &par9 MesonRadius_f(0)(980)
+    Value: 1.0
+    Fix: true
+    Type: MesonRadius
+  - Name: &par10 Position_f(0)(1500)
+    Value: 1.506
+  - Name: &par11 Width_f(0)(1500)
+    Value: 0.112
+  - Name: &par12 MesonRadius_f(0)(1500)
+    Value: 1.0
+    Fix: true
+    Type: MesonRadius
 
 Intensity:
   Class: StrengthIntensity
@@ -412,4 +428,20 @@ Dynamics:
     Type: NonDynamic
     FormFactor:
       Type: BlattWeisskopf
-      MesonRadius: 1.0
+      MesonRadius: *par6
+  f(0)(980):
+    Type: RelativisticBreitWigner
+    PoleParameters:
+      Real: *par7
+      Imaginary: *par8
+    FormFactor:
+      Type: BlattWeisskopf
+      MesonRadius: *par9
+  f(0)(1500):
+    Type: RelativisticBreitWigner
+    PoleParameters:
+      Real: *par10
+      Imaginary: *par11
+    FormFactor:
+      Type: BlattWeisskopf
+      MesonRadius: *par12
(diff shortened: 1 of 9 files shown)


tests/unit/amplitude/test_yaml_helicity.py
```diff
@@ -91,32 +91,27 @@ def test_kinematics_section(imported_dict):
...
 def test_dynamics_section(imported_dict):
     dynamics = imported_dict["Dynamics"]
-    assert len(dynamics) == 1
+    assert len(dynamics) == 3
 
     j_psi = dynamics["J/psi(1S)"]
     assert j_psi["Type"] == "NonDynamic"
     assert j_psi["FormFactor"]["Type"] == "BlattWeisskopf"
-    assert j_psi["FormFactor"]["MesonRadius"] == 1.0
+    assert j_psi["FormFactor"]["MesonRadius"] == "MesonRadius_J/psi(1S)"
```
spflueger, line 108: Test the value as well?

tests/unit/amplitude/expected_recipe.yml
```diff
@@ -28,10 +28,26 @@ Parameters:
...
     Value: 1.0
     Fix: true
+  - Name: &par7 Position_f(0)(980)
+    Value: 0.99
+  - Name: &par8 Width_f(0)(980)
+    Value: 0.06
+  - Name: &par9 MesonRadius_f(0)(980)
+    Value: 1.0
+    Fix: true
+    Type: MesonRadius
```
spflueger, line 42: Is the `Type` actually used/needed?

redeboer: Nope 😬 Remove it?

spflueger: Yes. Probably we also don't need it for the mag phase Type. But there we might have to put a little bit of thought into it, as we want to support also real and imaginary part

tests/unit/amplitude/test_yaml_helicity.py
spflueger: Afaiu that should be available through the fit parameters section now, using that name right?

tests/unit/amplitude/expected_recipe.yml
redeboer: That's true. I ask, because the current implementation does not hurt: it's not used, but we're not sure if it's needed.

_Note_: types are note provided for the mass and with (pole position) parameters

tests/unit/amplitude/test_yaml_helicity.py
redeboer: Sorry, overlooked this comment. Yes actually it was a bug: the meson radius from the fit parameters section should have been referenced, not a hard-coded value.

tests/unit/amplitude/expected_recipe.yml
spflueger: You can also leave it but we should open an issue then

tests/unit/amplitude/test_yaml_helicity.py
spflueger: so the additional test im proposing would make sure that the fit parameter is also registered correctly

tests/unit/amplitude/expected_recipe.yml
redeboer: #161

redeboer: https://github.com/ComPWA/tensorwaves/issues/161

spflueger: I opened a new issue for this #397

tests/unit/amplitude/test_yaml_helicity.py
spflueger: Well the tests need a lot more improvements anyway. So I leave it up to you if you want to add anything to this.

spflueger (approved):

tests/unit/amplitude/test_yaml_helicity.py
redeboer: ad7e2d7f
But I indeed doubt this test adds much

PR #401, opened 2020-12-02
refactor!: make solving module graph independent

### LINKED ISSUE #380: Change solver interface and use EdgeQuantumNumber instead of Particle
This allows to extract the conversion functionality `Particle` <--> `EdgeQuantumNumber` and easily reuse it across several solvers. Also decouples the particle and solving module.

spflueger:
- Change interface of Solver, to accept a new data structure ProblemSet
instead of the previous arguments. Also introduce a new data structure
QNResult which is now also the new return type of the Solver interface.
This removes the StateTransitionGraph and Particle from the module.
- Move Result to the reaction module to keep the interface for
find_solutions of the STM stable.
- Change prepare_graphs and underlying code to create_problem_sets. This
instead creates ProblemSet data structures that can be fed to the
solving module.
- Adapt other code pieces to these changes.

Closes #380 
Connects to #266

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -99,6 +99,7 @@
     "ignoreWords": [
         "adrs",
         "arange",
+        "asdict",
         "cano",
         "celltoolbar",
         "codacy",
--- a/src/expertsystem/reaction/__init__.py
+++ b/src/expertsystem/reaction/__init__.py
@@ -8,7 +8,8 @@ particle reaction problems.
 
 import logging
 import multiprocessing
-from copy import copy, deepcopy
+from collections import defaultdict
+from copy import deepcopy
 from enum import Enum, auto
 from itertools import product
 from multiprocessing import Pool
@@ -24,10 +25,11 @@ from typing import (
     Union,
 )
 
+import attr
 from tqdm import tqdm
 
 from expertsystem import io
-from expertsystem.particle import ParticleCollection
+from expertsystem.particle import Particle, ParticleCollection
 from expertsystem.reaction.conservation_rules import (
     BaryonNumberConservation,
     BottomnessConservation,
@@ -57,16 +59,19 @@ from ._default_settings import (
 from ._system_control import (
     CompareGraphNodePropertiesFunctor,
     GammaCheck,
-    GraphSettingsGroups,
     InteractionDeterminator,
+    InteractionTypes,
     LeptonCheck,
+    create_edge_properties,
+    create_interaction_properties,
+    create_node_properties,
+    create_particle,
     filter_interaction_types,
-    group_by_strength,
     remove_duplicate_solutions,
 )
 from .combinatorics import (
     StateDefinition,
-    initialize_graph,
+    create_initial_facts,
     match_external_edges,
 )
 from .quantum_numbers import (
@@ -78,12 +83,13 @@ from .quantum_numbers import (
 from .solving import (
     CSPSolver,
     EdgeSettings,
+    GraphElementProperties,
     GraphSettings,
-    InteractionTypes,
     NodeSettings,
-    Result,
+    ProblemSet,
+    QNResult,
     Rule,
-    validate_fully_initialized_graph,
+    validate_full_solution,
 )
 from .topology import (
     InteractionNode,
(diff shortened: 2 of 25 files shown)


spflueger:
@redeboer So the PR will not pass all of the tests yet. But all in all this is the solution i propose. Take a look and let me know what you think. Unfortunately a lot of modules where quite tightly coupled so that I had to touch quite a few code pieces to get this done.

src/expertsystem/reaction/topology.py
```diff
@@ -61,7 +61,7 @@ def __init__(
         if nodes is not None:
             self.__nodes = set(nodes)
         if edges is not None:
-            self.__edges = edges
+            self.__edges = dict(edges)
```
redeboer, line 64: May want to change the type hint into
```python
edges: Optional[Mapping[int, Edge]] = None
```
because the cast into `dict` allows you to give more general mapping structures as input (e.g. using `zip`).

src/expertsystem/reaction/topology.py
```diff
@@ -34,8 +34,8 @@
 class Edge:
     """Struct-like definition of an edge, used in `Topology`."""
 
-    ending_node_id: Optional[int] = attr.ib(default=None)
     originating_node_id: Optional[int] = attr.ib(default=None)
+    ending_node_id: Optional[int] = attr.ib(default=None)
```
redeboer, line 38: Note: I thought ending node was put first, because of common notation (think of _M\_fi_). I also remember there were some implementation issues where it was more convenient to put ending node first, because initial state is usually just one edge (can't remember where exactly).

src/expertsystem/reaction/topology.py
```diff
@@ -73,7 +73,7 @@ def edges(self) -> Dict[int, Edge]:
         return self.__edges
 
     def __repr__(self) -> str:
-        return f"{self.__class__.__name__}{(self.nodes, self.edges)}"
+        return f"{self.__class__.__name__}{(self.__nodes, self.__edges)}"
```
redeboer, line 76: I wonde what is best to do in these situations: call the internals of `self`, or use its interface (e.g. `self.nodes`, as happened previously). Same further down.

src/expertsystem/reaction/topology.py
```diff
@@ -315,19 +309,28 @@ class StateTransitionGraph(Topology, Generic[_EdgeType]):
 
     def __init__(
         self,
-        nodes: Optional[Set[int]] = None,
-        edges: Optional[Dict[int, Edge]] = None,
+        topology: Topology,
```
redeboer, line 312: I wonder if it's worth using `attr.s` here, now `StateTransitionGraph` is using composition only.

src/expertsystem/reaction/topology.py
```diff
@@ -336,42 +339,139 @@ def __eq__(self, other: object) -> bool:
...
+            return all(
+                self.get_node_props(i) == other.get_node_props(i)
+                for i in self.nodes
+            )
 
         raise NotImplementedError
 
-    def __copy__(self) -> "StateTransitionGraph[_EdgeType]":
-        """Makes a *shallow* copy.
+    @property
```
redeboer, line 361: Actually, why use composition?

src/expertsystem/reaction/topology.py
```diff
@@ -336,42 +339,139 @@ def __eq__(self, other: object) -> bool:
...
+            temp_edge_list = new_temp_edge_list
+        return edge_list
+
+    def get_node_props(self, node_id: int) -> InteractionProperties:
+        return self.__node_props[node_id]
+
+    def get_edge_props(self, edge_id: int) -> _EdgeType:
+        return self.__edge_props[edge_id]
+
+    def evolve(
```
redeboer, line 441: I like this new method. Might need a docstring here though, because it seems to be an important method.

src/expertsystem/reaction/topology.py
```diff
@@ -476,17 +576,6 @@ def build_graphs(
 
                 extendable_graph_list.extend(self.extend_graph(active_graph))
 
-            # check if two topologies are the same
```
redeboer, line 479: Why has this become redundant?

redeboer (commented):
(_Partial review, while @spflueger is designing a new `ProblemSet` class_)
Looks good so far, good job!
Here are some thoughts while reading through the changes to the `topology` sub-module.

src/expertsystem/reaction/topology.py
spflueger: It just seems counter intuitive to me. Yeah the amplitude M_fi is written "reverse", but I think that's a separate topic. When building a topology and I want to make a connection from node 0 to node 1 (note that the edges have direction), then I want to write this in the form `Edge(0,1)`. Currently it was `Edge(1,0)`.

src/expertsystem/reaction/topology.py
spflueger: That is true. I would say using the interface would be more consistent, so I will change that back.

src/expertsystem/reaction/topology.py
spflueger: Well the check was never performed since is isomorphic function is not implemented. Hence I just deleted unused code.

src/expertsystem/reaction/topology.py
spflueger: Could be. The main change here is that I wanted to stabilize the interface and make the node and edge props private. Theoretically the implementation underneath could change as long as it adapts to the interface at the end

src/expertsystem/reaction/topology.py
spflueger: You mean why not inherit from topology? If thats the question, the idea was to make the Graph "frozen" from the topology point of view. So the add_node and edge functionality is not there anymore.

redeboer: Yes, previously `StateTransitionGraph` inherited from `Topology`. But sounds good 👍

src/expertsystem/reaction/topology.py
redeboer: Ok. I initially also preferred originating node first, but I remember there was some reason not to do so. It's indeed better if originating node first works find in the current implementation.

src/expertsystem/reaction/solving.py
```diff
@@ -104,6 +90,35 @@ class GraphSettings:
...
+
+    initial_facts
+    solving_settings
+    """
+
+    topology: Topology = attr.ib()
+    initial_facts: GraphElementProperties = attr.ib()
+    solving_settings: GraphSettings = attr.ib()
+
+    # def __validate(self):
```
redeboer, line 111: Can be removed?

src/expertsystem/reaction/solving.py
```diff
@@ -104,6 +90,35 @@ class GraphSettings:
...
+class GraphElementProperties:
+    edge_props: Dict[int, GraphEdgePropertyMap] = attr.ib(factory=dict)
+    node_props: Dict[int, GraphNodePropertyMap] = attr.ib(factory=dict)
+
+
+@attr.s(frozen=True)
+class ProblemSet:
+    """Particle reaction problem set, defined as a graph like data structure.
+
+    initial_facts
```
redeboer, line 103: Maybe
```
Args:
    initial_fact (...): ...
```
?

src/expertsystem/reaction/solving.py
```diff
@@ -148,15 +163,13 @@ def get_name(rule: Any) -> str:
     return converted_dict
 
 
-class Result:
+class QNResult:
```
redeboer, line 166: The name is not that instructive. But I think eventually (as discussed elsewhere) this class needs to be split into a container of solutions and some bookkeeping clss for violated rules etc.

src/expertsystem/reaction/__init__.py
```diff
@@ -262,87 +550,86 @@ def set_allowed_interaction_types(
...
+        initial_facts = create_initial_facts(
+            topology=topology,
+            particles=self.__particles,
+            initial_state=self.initial_state,
+            final_state=self.final_state,
+            final_state_groupings=self.final_state_groupings,
+        )
+
+        logging.info(f"initialized {len(initial_facts)} graphs!")
+        return [InitialFacts(edge_props=x) for x in initial_facts]
```
redeboer, line 594: Shouldn't `create_initial_facts` return an instance of `InitialFacts`? In other words, why is `InitialFacts` defined under `reaction.__init__` and not under `combinatorics`?

src/expertsystem/reaction/solving.py
spflueger: I can also rename those to `QuantumNumberResult` result.

src/expertsystem/reaction/topology.py
spflueger: done

src/expertsystem/reaction/topology.py
spflueger: done

src/expertsystem/reaction/solving.py
spflueger: done

src/expertsystem/reaction/__init__.py
spflueger: Good point. e263774e6ad99f0965b734037c37bf99ec905449

src/expertsystem/reaction/solving.py
spflueger: removed

tests/unit/reaction/test_combinatorics.py
```diff
@@ -101,14 +99,13 @@ def test_constructor():
...
-    def test_from_graph(three_body_decay, particle_database):
-        graph1 = StateTransitionGraph.from_topology(three_body_decay)
-        pion = particle_database["pi0"]
-        graph1.edge_props[0] = ("J/psi", [-1, +1])
-        graph1.edge_props[2] = pion
-        graph1.edge_props[3] = pion
-        graph1.edge_props[4] = "gamma"
-        kinematic_representation1 = _get_kinematic_representation(graph1)
+    def test_from_topology(three_body_decay):
+        pi0 = ("pi0", [0])
```
redeboer, line 103: Is it required to specify the spin projections now?

src/expertsystem/reaction/_system_control.py
```diff
@@ -26,6 +30,103 @@
...
+    spin_projection: Optional[float] = None,
+) -> GraphEdgePropertyMap:
+    edge_qn_mapping: Dict[str, Type[EdgeQuantumNumber]] = {
+        qn_name: qn_type
+        for qn_name, qn_type in EdgeQuantumNumbers.__dict__.items()
+        if not qn_name.startswith("__")
+    }  # Note using attr.fields does not work here because init=False
+    property_map: GraphEdgePropertyMap = {}
+    isospin = None
+    for qn_name, value in attr.asdict(particle).items():
```
redeboer, line 44: Isn't it better to iterate with `attr.fields`?

src/expertsystem/reaction/_system_control.py
```diff
@@ -26,6 +30,103 @@
...
+def create_node_properties(
+    node_props: InteractionProperties,
+) -> GraphNodePropertyMap:
+    node_qn_mapping: Dict[str, Type[NodeQuantumNumber]] = {
+        qn_name: qn_type
+        for qn_name, qn_type in NodeQuantumNumbers.__dict__.items()
+        if not qn_name.startswith("__")
+    }  # Note using attr.fields does not work here because init=False
+    property_map: GraphNodePropertyMap = {}
+    for qn_name, value in attr.asdict(node_props).items():
```
redeboer, line 74: Same here

cspell.json
```diff
@@ -99,6 +99,7 @@
     "ignoreWords": [
         "adrs",
         "arange",
+        "asdict",
```
redeboer, line 102: See remark on `attr.fields` -- this line can be removed if true

redeboer (approved):
Good job! Just a few remarks

src/expertsystem/reaction/_system_control.py
spflueger: See the code comment, the `attr.fields` are empty (i believe because of `init=False`)

redeboer: Ah yes overlooked that

tests/unit/reaction/test_combinatorics.py
spflueger: was that different before?

redeboer: I would have expected `"pi0"` would suffice. But actually the test has become a bit different: it now uses this tuple in the kinematic representation, previously it was an instance of `Particle`. But here, both are fine, they are just used to compare the permutations.

PR #402, opened 2020-12-03
ci: update requirements through GitHub Actions

### LINKED ISSUE #409: Pin all nested dependencies for different Python versions
_This issue serves as a summary of the findings for #402_

Since #102, we list all dependencies in the [`setup.cfg`](https://github.com/ComPWA/expertsystem/blob/7daef69533b19e7c57122bb23fe44bedfe429bc2/setup.cfg) file. We make a distinction between package requirements (defined by `install_requires`) and additional dependencies (defined through different sections like `doc` and `sty` under `extras_require`).

Each of the versions of dependencies were pinned in the `setup.cfg` file, so that developers use the same version. However, this does not pin _indirect_ dependencies, like `pyrsistent` (which is required by `jsonschema`)

Since #394, we therefore pin _both direct and indirect_ dependencies through `requirements.txt` and `requirements-dev.txt`. These files are created with `pip-compile` ([`pip-tools`](https://github.com/jazzband/pip-tools)) in a `tox` job (`tox -e upgrade`). Input files are **unpinned** dependency files (in **bold** the files that are relevant for the user/developer):

| Unpinned        | Pinned output              |
| :-------------- | :------------------------- |
| **`setup.cfg`** | **`requirements.txt`**     |
| `.reqs/doc.in`  | `.reqs/doc.txt`            |
| `.reqs/sty.in`  | `.reqs/sty.txt`            |
| `.reqs/test.in` | `.reqs/test.txt`           |
| `.reqs/dev.in`  | **`requirements-dev.txt`** |

There is a certain hierarchy in these files: for instance, `requirements-dev.txt` contains all dependencies that are listed in `requirements.txt`. This dependency is defined in the `tox -e upgrade` job.

There are some problems/challenges with this setup though:

1. Requirements differ per Python version. For instance, some package work with `importlib-resources` for Python <= 3.7. There should therefore be versions for each of the files in the right column _for each version of Python_.

   > **Proposed solution**
   >
   > Follow a new folder structure: remove the `.reqs` folder in favour of a `dev` folder that contains the [`.in` files](https://github.com/jazzband/pip-tools#example-usage-for-pip-compile). These files list _unpinned_ dependencies (at most a certain version range). This `dev` folder has subfolders `dev/3.6`, `dev/3.7`, and `dev/3.8` that contain _pinned_ requirement files. These files are created with `pip-compile` from the `.in` files.

2. Ideally, we want these pinned requirement files to be compatible, so that you can for instance do

   ```bash
   pip install -r requirements.txt -r requirements-dev.txt
   ```

   In other words, there should be no version conflicts between duplicate dependencies listed in each of the files.

   > **Proposed solution**
   >
   > Run a specific sequence of `pip-compile` commands through a GitHub Action (see `ci-dev.yml`). This allows us to created pinned `.txt` files first, then use those as constraint files (`-c`) in the `.in` files. So there will be two types of `.in` files: the ones that only list the dependencies themselves (for instance `*-sty.in`) and the ones that the ones that use those `.in` files along with constraints/requirements (`-c`/`-r`), say, `*-sty-deps.in`.
   >
   > Splitting the `.in` files into 'bare' requirement listings and dependency listings between the `.in` files is required for 'forward constraining', that is ensuring that `pip-compile` takes into account that, say, `requirements-sty.in` will introduce a version of `importlib-metadata` that is limited to a certain max, while the (prior) `requirements-test.in` will pin `importlib-metadata` to the latest release. This would cause conflicts when creating `requirements-dev.txt`, which should contain both `requirements-test.txt` and `requirements-sty.txt`.
   >
   > **One problem here: `pip-compile` does not support constraint flags, so we have to feed these through the `*.in` files. This also means we need a `requirements.in` files that lists exactly the same as the `setup.cfg` file.**

3. All `requirement*.txt` files need to be upgraded once one of the dependencies changes. Dependency changes only take place in `setup.cfg` and the `*.in` files.

   > **Proposed solution**
   >
   > The `requirements*.txt` files should be updated automatically, so this has to be done through a GitHub Action _upon every PR_. GitHub Actions cannot push to branches that are on a fork, so this means a separate PR should be spawned _if_ there changed in the `setup.cfg` or `.in` files are detected.

4. How to handle updates in the dependencies?

   > In order to keep the pinned `requirements*.txt` files up-to-date, a cron job should _upgrade_ these files. This is the same as 3., but using `pip-compile` with the flag `--upgrade`.

redeboer:
Closes #409 

Requirements differ for each version of Python. We therefore have to maintain `requirement.txt` files for each version of Python that we supports. It is, however, hard to update these lists manually, so it is better to do this through a GitHub workflow (which has access to all versions of Python).

See also
https://github.com/jazzband/pip-tools#cross-environment-usage-of-requirementsinrequirementstxt-and-pip-compile 

- [x] Write `requirements.txt` files to different subfolders `dev/3.6`, `dev/3.7`, `dev/3.8` in GitHub Actions
- [x] Fix dependency conflicts `importlib-resources` and `importlib-metadata` for Python 3.6-3.7
- [x] Run only on changes to `setup.cfg` and `*.in` files (`git diff` non-empty) --> abac046 and f17f0d8
- [x] Revert to using `skip-install` and [`whitelist_externals`](https://tox.readthedocs.io/en/latest/config.html?highlight=whitelist_externals#conf-allowlist_externals) in `tox` --> 8c2c1e0
- [x] Pin Python version in Conda `environment.yml` file --> 7f2a06c
- [x] Re-introduce `tox -e upgrade` --> 0600b01
- [x] Commit and push changes to PR through GitHub Actions upon changes to requirements --> https://github.com/ComPWA/expertsystem/pull/402#issuecomment-746171501
- [x] Run cron job for master branch --> 0c605ee
- [x] Test the cron job in a temporary repository --> https://github.com/ComPWA/expertsystem/pull/402#issuecomment-746171501


redeboer:
@spflueger f6561d7 sketches the ideas I formulated in #409. However, the idea of 'forward constraints' ― see e.g.

https://github.com/ComPWA/expertsystem/blob/f6561d7cb4199f1223ca453c9e66dae228285bab/dev/hierarchy/requirements-base.in#L3-L7

does not work currently.

redeboer:
Update: 104b7f4 also won't work, but I pushed it for illustration

spflueger:
> ./requirements-doc.in

Don't the constraints have to be `.txt` files since we want to have pinned versions there?

redeboer:
> Don't the constraints have to be .txt files since we want to have pinned versions there?

Ideally yes, but they don't exist yet at this stage.

spflueger:
> > Don't the constraints have to be .txt files since we want to have pinned versions there?
> 
> Ideally yes, but they don't exist yet at this stage.

So I thought the process is the following:
1. create the `req.txt` of the full env with `pip-compile`
2. use that as the constraint for all of the sub envs

redeboer:
> So I thought the process is the following: create the req.txt of the full env with pip-compile using all of the .in files use that req.txt as the constraint for all of the sub envs

Yeah I'm working on that setup too. Currently, the constraints are specified using the `.in` files ('forward constraints'), but those constraints don't work recursively into the indirect dependencies. So we would want to span all allowed packages (the `dev` requirements in our case) through a pinned dependencies file (`.txt`).

Now the problem is that afaics, `pip-compile` cannot specify constraints through some flag, only through `pip-requirement` syntax. The pinned `.txt` are written to subfolders specific to the python version though.

Trying to find a way around it...

redeboer:
https://github.com/ComPWA/expertsystem/pull/402#issuecomment-745481677
fb58433 should do the trick. Quite a monster, but it's the best I can get out of `pip-tools` for now.

.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r .reqs/test.txt
+          pip install -r dev/${{ matrix.python-version }}/requirements-test.txt
```
spflueger, line 34: For me the dirname `dev` is a bit misleading. `env(s)` or `req(s)` sounds better to me. What was the intention behind `dev`?

.github/workflows/requirements.yml
```diff
@@ -0,0 +1,66 @@
+# cspell:ignore noreply
+
+name: Upgrade requirements
+
+on:
+  schedule:
+    - cron: "0 2 * * 1"
```
spflueger, line 7: We can think about running this every 2 days, since it wont actually do anything if there are no changes. That way we cover more of the package combinations that exist and can achieve a bit higher compatibility.

dev/upgrade.sh
```diff
@@ -0,0 +1,24 @@
...
+mkdir -p dev/$PYTHON_VERSION &&
+    python dev/extract_install_requires.py &&
+    cp dev/requirements*.in dev/$PYTHON_VERSION/ &&
+    rm dev/$PYTHON_VERSION/requirements-dev.in &&
+    pip-compile --upgrade \
+        dev/requirements*.in \
+        -o dev/$PYTHON_VERSION/requirements-dev.txt &&
+    for in_file in $(ls dev/$PYTHON_VERSION/requirements*.in); do
+        echo -e "-c requirements-dev.txt\n$(cat ${in_file})" >${in_file}
+        pip-compile "${in_file}"
```
spflueger, line 19: You could try using a pipe `|` here. So something like
```
echo -e "-c requirements-dev.txt\n$(cat ${in_file})" | pip-compile
```

.github/workflows/requirements.yml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  requirements:
+    name: Upgrade requirement files
```
spflueger, line 11: This only checks out the default branch right?

.github/workflows/requirements.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@master
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pip-tools
+      - name: Upgrade dependencies
+        run: bash dev/upgrade.sh
```
spflueger, line 31: So this will return an error code when there are no changes, right? So any actual error in `upgrade.sh` will be "skipped".

It would be better if we could decide whether to make a new PR based on a diff of the requirement files.

.github/workflows/requirements.yml
```diff
@@ -0,0 +1,66 @@
+# cspell:ignore noreply
+
+name: Upgrade requirements
+
+on:
```
spflueger, line 5: This should also run on PRs (that change the setup `install_requires`), right?

spflueger (commented):
Nice work! Almost there :+1:

.github/workflows/requirements.yml
spflueger: Ok, I just saw that this is still on the todo list. So nvm.

.github/workflows/ci.yml
redeboer: Yes that's true. Initially I thought to put all things related to `dev` in there. Ideally that would also mean all these config files like `.pylintrc`, but that's not how these tools work (unless you specify every config file with a flag). For the same reason, I also initially named the workflow `ci-dev.yml`.
I'll go for `req`, I think. Some tools use `requirements` (e.g. [`pip-tools-multi`](https://pypi.org/project/pip-compile-multi/)), but then the full paths become rather repetitive.

.github/workflows/requirements.yml
redeboer: Considering the large number of packages in `requirements-dev.txt`, I'm afraid this would result in an upgrade PR every two days. Just one of those indirect dependencies would have to release an upgrade and the cron job would create a new PR.

.github/workflows/requirements.yml
redeboer: No, it will always upgrade. But if there are no file changes (in this cron job), there is nothing to commit, and the push job afterwards won't create a PR.

dev/upgrade.sh
redeboer: Yeah true, but I thought it (theoretically) more transparent to write this line to the output `.in` file, even if it is not committed.

.github/workflows/requirements.yml
redeboer: Seems like it, yes

> The branch, tag or SHA to checkout. When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. _Otherwise, uses the default branch._

https://github.com/actions/checkout#usage

.github/workflows/requirements.yml
spflueger: Hmm, but I think this would help us. Rather have many small updates than one big one. So in case there is some compatibility issue, we can easier figure out which package is causing it.

.github/workflows/ci.yml
redeboer: 1c75aa2

.github/workflows/requirements.yml
redeboer: Ok can change it. I just thought it might clog the commit history, see e.g.
https://github.com/peterdemin/pip-compile-multi/commits/master

dev/upgrade.sh
spflueger: To me using the stdin/out with pipes is much more elegant. What confused me is that this `in_file` is overwritten. So was not that transparent to me :grimacing:

.github/workflows/requirements.yml
spflueger: So then why is there a `continue-on-error` in the workflow?

redeboer: There is no error (I assume you refer to the `exit` codes in the `upgrade.sh` script). Unless it's something fatal like `pip-compile` not existing or being unable to determine the Python version.

redeboer: Ah ok `continue-on-error` is there so that the 3.6, 3.7 and 3.8 are all run if there is some problem. I did that for debugging purposes when there were still conflicts in the `txt` files.

.github/workflows/requirements.yml
spflueger: I see. We can always reduce it, if it would be too annoying

.github/workflows/requirements.yml
spflueger: The problem I have now is that the workflow setting `continue-on-error` will just skip this build if there are errors and not really pop up as red, probably as gray (maybe green). So I don't understand why `continue-on-error` is there/

dev/upgrade.sh
redeboer: This is the way to prepend to a file in bash. The line has to be prepended (not `>>`), because the`pip` constraints have to be specified first.

.github/workflows/requirements.yml
redeboer: GitHub kills the other jobs in the matrix is one of them fails. But you're right, the push job would still run if there are some problems that still result in are changes to commit. 65198e3 should take care of it, but perhaps it's indeed better to remove this `continue-on-error`.

redeboer: Done 925f344 :heavy_check_mark:

dev/upgrade.sh
redeboer: > To me using the stdin/out with pipes is much more elegant.

Note that the files have to be written to the `3.6`, `3.7`, `3.8` subfolders, because there are relative path references to the `.txt` files. These files are in the subfolders, not the main `reqs` folder.

spflueger: I know, well you are "prepending" by overwriting the file completely. I this actually what we want? That in file should stay untouched right?
Anyways the pipe would be much more transparent to me

redeboer: The `in_file` is a copy.
https://github.com/ComPWA/expertsystem/blob/925f344512745b03a055531268784762aecbdb4e/reqs/upgrade.sh#L12

I just tried with a pipe, but the problem remains that `pip-compile` needs to know which `requirements-dev.txt` file to use as constraints. For that it needs to know which subfolder to use, and since `pip-compile` does not support the `-c` constraint flag, this needs to be done in an `.in` file _that is placed in that subfolder_ (so that it can refer to the `requirements-dev.txt` that has been dumped to that folder previously).

spflueger: Ok maybe add a comment that `pip-compile` does not support this, so that no one else stumbles there

redeboer:
Some updates:
- e323ad0 introduces a second workflow that upgrades dependencies. So there are now two (almost similar) workflows that upgrade the developer dependencies:
  1. One that does this *only* for PRs and only if there are changes to `setup.cfg` or the `reqs` folder. If there are changes to the output `requirement*.txt` files, a new commit is pushed to the branch that triggered the PR. For this reason, **dependencies as defined through `setup.cfg` and the `reqs` folder can only be modified through a branch PR, not a fork PR**.
  2. One cron job that always performs `pip-compile --upgrade`. This job creates a new PR to the master branch.
- The cron job works correctly: https://github.com/redeboer/es-test/actions/runs/425532557. However, there seem to be access problems when creating the PR.
- The PR job works correctly: https://github.com/redeboer/es-test/pull/1/commits/2fbd5c98c325e3fb539968aad8d8adef818c39b2 was created through https://github.com/redeboer/es-test/runs/1563343873. Still to do: (1) the push job should not run if the PR is created from a fork and (2) the workflow should fail if there are changes to the dependencies from a fork PR.

redeboer:
Closing this PR in favour of #412, which comes from a branch on this repository (not a fork).

PR #417, opened 2021-01-06
fix: officially remove top NormalizedIntensity

spflueger:
Currently the expertsystem and tensorwaves are not compatible, as the highest level `NormalizedIntensity` is missing but is required when performing fits in tensorwaves.

To simplify things I propose the following:
We officially remove this highest level `NormalizedIntensity` as this is only required by fitting using a log likelihood estimator and belong there. This has several advantages:
1. It makes the amplitude model simpler as the normalization is not needed for the intensity definition
2. Data generation is fast as the normalization is skipped
3. The construction of the intensity in tensorwaves becomes more transparent
    ```python3
    builder = IntensityBuilder(part_list, kinematics) # note: the phase space sample here is optional
    intensity = builder.create_intensity(model)

    data_sample = generate_data(10000, kinematics, intensity)
    dataset = kinematics.convert(data_sample)

    phsp_sample = generate_phsp(100000, kinematics, random_generator, bunch_size=100)
    phspset = kinematics.convert(phsp_sample)

    estimator = UnbinnedNLL(intensity, dataset, phspset) # phspset is required here for the normalization
    ```
@redeboer @Leongrim @sebastianJaeger What do you think?

### DIFF
--- a/src/expertsystem/amplitude/helicity_decay.py
+++ b/src/expertsystem/amplitude/helicity_decay.py
@@ -25,7 +25,6 @@ from .model import (
     IncoherentIntensity,
     IntensityNode,
     Kinematics,
-    NormalizedIntensity,
     ParticleDynamics,
     RecoilSystem,
     SequentialAmplitude,
@@ -444,14 +443,11 @@ class HelicityAmplitudeGenerator:
                 graph_group
             )
             incoherent_intensity.intensities.append(coherent_intensity)
-        if len(incoherent_intensity.intensities) > 1:
-            intensity: IntensityNode = NormalizedIntensity(
-                incoherent_intensity
-            )
-        else:
+        if len(incoherent_intensity.intensities) == 1:
             intensity = incoherent_intensity.intensities[0]
-        strength_intensity = self.__prepend_strength(intensity)
-        return strength_intensity
+        else:
+            intensity = incoherent_intensity
+        return intensity
 
     def __create_parameter_couplings(
         self, graph_groups: List[List[StateTransitionGraph[ParticleWithSpin]]]
--- a/src/expertsystem/amplitude/helicity_decay.py
+++ b/src/expertsystem/amplitude/helicity_decay.py
@@ -442,11 +442,11 @@ class HelicityAmplitudeGenerator:
                 graph_group
             )
             incoherent_intensity.intensities.append(coherent_intensity)
+        if len(incoherent_intensity.intensities) == 0:
+            raise ValueError("List of incoherent intensities cannot be empty")
         if len(incoherent_intensity.intensities) == 1:
-            intensity = incoherent_intensity.intensities[0]
-        else:
-            intensity = incoherent_intensity
-        return intensity
+            return incoherent_intensity.intensities[0]
+        return incoherent_intensity
 
     def __create_parameter_couplings(
         self, graph_groups: List[List[StateTransitionGraph[ParticleWithSpin]]]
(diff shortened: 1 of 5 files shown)


redeboer:
Just a small note I can't put in the review: can `__prepend_strength` be removed?

https://github.com/ComPWA/expertsystem/blob/79e59f33a9cab970abc20073d15f742bcd776fa6/src/expertsystem/amplitude/helicity_decay.py#L477-L489

redeboer:
But generally I agree.

> 3. The construction of the intensity in tensorwaves becomes more transparent

Is that because a phase space sample is not required anymore?

redeboer:
I'll fix the tests now

spflueger:
> Just a small note I can't put in the review: can `__prepend_strength` be removed?
> 
> https://github.com/ComPWA/expertsystem/blob/79e59f33a9cab970abc20073d15f742bcd776fa6/src/expertsystem/amplitude/helicity_decay.py#L477-L489

It can be removed for now. It will be needed when using for example background parts in the amplitude model. So something like 
```
Intensity = strength * SignalIntensity + (1-strength) * BackgroundIntensity
```

redeboer:
https://github.com/ComPWA/expertsystem/pull/417#issuecomment-755369730
Ok. Because next question I would have is whether the `StrenghtIntensity` and `NormalizedIntensity` classes are still needed at all.

spflueger:
> But generally I agree.
> 
> > 1. The construction of the intensity in tensorwaves becomes more transparent
> 
> Is that because a phase space sample is not required anymore?

It **may** still be needed in the the `IntensityBuilder`, if normalizations within the amplitude model appear. For instance when the model includes
- signal and a background term. Then it can be beneficial to normalize both parts to have meaningful strengths (value E [0,1])
- normalizing amplitudes. Same logic as above. Magnitudes then correspond to fit fractions.
- maybe more use cases

src/expertsystem/amplitude/helicity_decay.py
```diff
@@ -444,14 +442,11 @@ def __generate_intensities(
                 graph_group
             )
             incoherent_intensity.intensities.append(coherent_intensity)
-        if len(incoherent_intensity.intensities) > 1:
-            intensity: IntensityNode = NormalizedIntensity(
-                incoherent_intensity
-            )
-        else:
+        if len(incoherent_intensity.intensities) == 1:
```
redeboer, line 445: I'm thinking what happens in the case when the `len` is 0. In that sense, the following might be better:
```python
if len(incoherent_intensity.intensities) == 0:
    raise ValueError("List of incoherent intensities cannot be empty")
if len(incoherent_intensity.intensities) == 1:
    return incoherent_intensity.intensities[0]
return incoherent_intensity
```
(where the first `if` statement could be removed if that is a redundant check)

redeboer (approved):
LGTM!
Just one comment about the if else statement

spflueger:
> [#417 (comment)](https://github.com/ComPWA/expertsystem/pull/417#issuecomment-755369730)
> Ok. Because next question I would have is whether the `StrenghtIntensity` and `NormalizedIntensity` classes are still needed at all.

See https://github.com/ComPWA/expertsystem/pull/417#issuecomment-755373859

src/expertsystem/amplitude/helicity_decay.py
spflueger: Good point

redeboer: d4aaaec

PR #421, opened 2021-01-06
fix: decouple parity violating amplitudes

### LINKED ISSUE #419: Parity violation in weak decays is not considered in coupling of amplitudes
When generating an Amplitude Model for the decay Lambda_c^+ -> proton K- pi+, the magnitudes and phases of weak decays are automatically coupled due to parity, although parity violation in weak decays means that they should be independent of each other. 

For example in the subdecay Lambda_c^+ -> proton K*(892), the parameter
```python
"Magnitude_Lambda(c)+_to_K*(892)0_1+p_0.5;K*(892)0_to_K-_0+pi+_0"
```
gets used in both the 
```python
"Lambda(c)+_-0.5_to_K*(892)0_-1+p_-0.5;K*(892)0_-1_to_K-_0+pi+_0"
```
and the
```python
"Lambda(c)+_-0.5_to_K*(892)0_1+p_0.5;K*(892)0_1_to_K-_0+pi+_0"
```
amplitude, though these should be independend of each other.

spflueger:
Decouple coefficients of amplitudes which are parity violating, by generating the correct coefficient names.

Closes #419 

- [x] Add test to verify decoupling for parity violating amplitudes

### DIFF
--- a/src/expertsystem/amplitude/helicity_decay.py
+++ b/src/expertsystem/amplitude/helicity_decay.py
@@ -280,6 +280,9 @@ class _HelicityAmplitudeNameGenerator:
                 priority_partner_coefficient_suffix,
             ) = self._generate_amplitude_coefficient_couple(graph, node_id)
 
+            if graph.get_node_props(node_id).parity_prefactor is None:
+                continue
+
             if (
                 coefficient_suffix
                 not in self.parity_partner_coefficient_mapping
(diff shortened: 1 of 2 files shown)


spflueger:
@sebastianJaeger Could you verify that this creates the correct amplitude coefficient couplings. I already checked it but just to verify.

redeboer:
> [ ] Add test to verify coupling for parity conserving amplitudes
> [ ] Add test to verify decoupling for parity violating amplitudes

Simplest would be to add some new channel test (`tests/channel`), for instance @sebastianJaeger's channel (versus what?). Better would be a proper unit test where you don't need to run the whole `StateTransitionManager`, but that is difficult to do.

redeboer (approved):
Good job!
I'm only thinking about the tests. Currently, only the expected number of parameters is tested. What can happen at most then is that the number changes and you have to figure out whether that change in number makes sense (could come from anything, e.g. #417 would have triggered that). So best would be to somehow compare a channel that does couple the parameters with one that don't, even if by expected number of parameters.
But this argument goes for many of the tests. Many of them don't really test for any logic, but instead just check whether the framework produces the right numbers.

Issue #385, opened 2020-11-13
Make serialization to dict public


redeboer:
Currently, the `expertsystem` provides write and load functions that read from a file and it's not possible to feed a dictionary directly. This is the reason why tests fail haphazardly on GitHub Actions: the tests have to write to disk in order to test the `io` module, which causes crashes when running tests in parallel.

redeboer:
This issue needs to be tackled to keep the CI/CD stable when running pytest in parallel mode: it's safer not to use fixtures in combination with read-write (here: reading and writing from and to YAML)

See e.g. this log:
https://github.com/ComPWA/expertsystem/runs/1460495658?check_suite_focus=true

Issue #384, opened 2020-11-13
Implement Serializable interface with asdict and fromdict


redeboer:
The original idea was to use a **function** (not a method) like [`attr.asdict`](https://www.attrs.org/en/stable/api.html#attr.asdict) (with `recurse=True`) to recursively serialize any data structure into a `dict`. Data structures that are not constructed through an `attr.s` constructor, can be serialized by defining a 'serializer', see https://www.attrs.org/en/stable/examples.html#converting-to-collections-types.

There are two big problems with this implementation:

1. We also want to construct the original data structures back from any `dict` that was created through this `asdict` function. If `AttrObject` is some class that is constructed with an `attr.s` decorator, this can be done as follows:
  ```python
  output_dict = attr.asdict(some_attr_object)
  AttrObject(**output_dict)
  ```
  However, this only works if `AttrObject` is not a nested data structure.

2. Classes such as `ParticleCollection`, `Spin`, and `Parity` are not created through `attr.s`. We would have to provide some serializer for them, but this still doesn't really address #383, as we would have to put that serializer under some submodule other than. See [here](https://www.attrs.org/en/stable/examples.html#converting-to-collections-types) for the syntax that would be required.

Conclusion: it would be better to define some `Protocol` or `ABC` that defines `fromdict` and `asdict` **methods** for any class that we want to be serializable, _including classes that were decorated with `attr.s`.

<details>
<summary>Some snippets of the <b>function</b> implementation</summary>

```python
# expertsystem 0.6.4

from typing import Any, Callable, Optional

import attr

import expertsystem as es
from expertsystem.particle import Parity, ParticleCollection, Spin

Serializer = Callable[[type, attr.Attribute, Any], Any]
"""https://www.attrs.org/en/stable/extending.html#customize-value-serialization-in-asdict"""


def asdict(
    instance: object,
    serializer: Optional[Serializer] = None,
) -> dict:
    if serializer is None:
        if not attr.has(type(instance)):
            serializer = value_serializer
    if not attr.has(type(instance)):
        return value_serializer("", "", instance)
    return attr.asdict(instance, recurse=True, value_serializer=serializer)


def value_serializer(_: type, __: attr.Attribute, value: Any) -> Any:
    if isinstance(value, ParticleCollection):
        return {
            "particles": [
                attr.asdict(p, value_serializer=value_serializer)
                for p in value
            ]
        }
    if isinstance(value, Parity):
        return {"value": int(value)}
[...]

PR #425, opened 2021-01-08
refactor!: remove XML support

redeboer:
Keeping up support for the old ComPWA format is becoming too bothersome with regard to #384 and #385. Eventually, we have to abandon XML support anyway, because we want to dump to and from `dict` structures and leave the file format choice (e.g. YAML or JSON) up to the user.


spflueger (approved):
:+1:

PR #436, opened 2021-01-11
refactor!: use attr.asdict in io module

redeboer:
The `io` module now uses `attr.asdict` where possible. This has a few consequences:
1. classes that are constructed with `attr.s` can be refactored without having to worry about the `io` module.
2. the amplitude model file changes along with the structure of the data classes
3. the code of the `io` module becomes simpler. This is a step towards removing this module (see e.g. #400)

## To-do list

- [x] `ParticleCollection`
- [x] `FitParameters`
- [x] `Dynamics`
- [x] `IntensityNode`
- [x] `Kinematics` (not an `attr.s` class)

### DIFF
--- a/src/expertsystem/io/_dict/dump.py
+++ b/src/expertsystem/io/_dict/dump.py
@@ -29,7 +29,7 @@ def from_amplitude_model(model: AmplitudeModel) -> dict:
     output_dict = {
         "Kinematics": __kinematics_to_dict(model.kinematics),
         **from_fit_parameters(model.parameters),
-        "Intensity": __intensity_to_dict(model.intensity),
+        "intensity": __intensity_to_dict(model.intensity),
         **from_particle_collection(model.particles),
         "dynamics": __dynamics_section_to_dict(model.dynamics),
     }
--- a/src/expertsystem/io/_dict/dump.py
+++ b/src/expertsystem/io/_dict/dump.py
@@ -57,7 +57,11 @@ def from_fit_parameters(parameters: FitParameters) -> dict:
 
 
 def from_fit_parameter(parameter: FitParameter) -> dict:
-    return attr.asdict(parameter, recurse=True)
+    return attr.asdict(
+        parameter,
+        recurse=True,
+        filter=lambda attr, value: attr.default != value,
+    )
 
 
 def __kinematics_to_dict(kin: Kinematics) -> dict:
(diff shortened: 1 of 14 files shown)


tests/unit/io/expected_recipe.yml
```diff
@@ -11,426 +11,434 @@ Kinematics:
...
-    Type: RelativisticBreitWigner
-    PoleParameters:
-      Real: *par6
-      Imaginary: *par7
-    FormFactor:
-      Type: BlattWeisskopf
-      MesonRadius: *par8
+    type: RelativisticBreitWigner
+    pole_position: *par6
+    pole_width: *par7
```
spflueger, line 434: We should really define a better name for the pole. The pole position in literature is the complex number, which defines the pole in the complex energy plane. Maybe we should name it `pole_real`, `pole_imag`.

tests/unit/test_io.py
```diff
@@ -24,6 +24,10 @@ def test_serialization(
     io.write(exported, filename)
     asdict = io.asdict(exported)
     imported = io.fromdict(asdict)
+    io.write(
+        imported,
+        output_dir + f"test_write_read_{formalism}.imported.{file_extension}",
+    )
```
spflueger, line 30: That's a separate test for testing the write functionality? Maybe this can be put into a separate test. To keep the tests more distinct

src/expertsystem/io/_dict/dump.py
```diff
@@ -135,136 +76,97 @@ def __kinematics_to_dict(kin: Kinematics) -> dict:
...
+        "type": node.__class__.__name__,
+        **attr.asdict(
+            node,
+            filter=lambda field, value: field.name
+            not in [
+                "amplitudes",
+                "amplitude",
+                "intensities",
+                "intensity",
+            ]
```
spflueger, line 108: I don't get this part. I was expecting some "simple" recursion

spflueger (commented):
Good job! Looks good. I just added some comments and a question, but more for my understanding

tests/unit/test_io.py
redeboer: No, I put write statements in there so you can still view the output in case there's a problem. The writing itself doesn't really need to be tested, because that would essentially be testing the `json` or `yaml` package (or, whether the nested `dict` can be dumped to those formats).

redeboer: Ok actually that line was some outdated code, fixed by ec8c489

src/expertsystem/io/_dict/dump.py
redeboer: Recursion works again, see 70e7d2e

Issue #440, opened 2021-01-12
Allow inserting custom dynamics


redeboer:
Since #434, it has become possible to do something like:
```python
@attr.s
class CustomDynamics(Dynamics):
    par1: FitParameter = attr.ib()
    par2: FitParameter = attr.ib()
    ...

custom_dynamics = CustomDynamics(...)
# model: AmplitudeModel
model.dynamics["J/psi(1S)"] = custom_dynamics
```
without breaking `asdict` and `fromdict` (since #436) and also inserting `par1` and `par2` into `model.parameters` (the `FitParameters` class that keeps track of which `FitParameter`s are in the `AmplitudeModel`. There's at least one bug though: if `"J/psi(1S)"` was already assigned some `Dynamics` instance with `FitParameter`s, the old `FitParameter`s are still in `model.parameters`. So the `FitParameters` class has to become mutable (see also #382).

Other bugs may be pointed out as well once tests are written and the procedure illustrated in the notebooks.

spflueger:
Is overhauled by PR #454

PR #443, opened 2021-01-12
feat: add functionality to couple fit parameters

### LINKED ISSUE #382: Add functionality to couple fit parameters
The [`FitParameters`](https://pwa.readthedocs.io/projects/expertsystem/en/0.6.3/api/expertsystem.amplitude.model.html#expertsystem.amplitude.model.FitParameters) class is currently rather (too) simple: it only provides functionality to search for parameters and to add them. This is a problem if one wants to couple fit parameters.

A simple solution would be to make have the [`FitParameters`](https://pwa.readthedocs.io/projects/expertsystem/en/0.6.3/api/expertsystem.amplitude.model.html#expertsystem.amplitude.model.FitParameters) derive from a [`MutableMapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableMapping) and implement the required methods. One can then do something like:

```python
model.parameters["Width_N(1720)~-"] = model.parameters["Width_N(1720)+"]
```

which is, I think (@spflueger?), sufficient to couple those parameters in `tensorwaves`.

A next step would of course be to build some more intelligence in the model building. But there we have to be careful not to tailor too much to specific analysis cases.

redeboer:
Closes #382 

_Rebased on #442_

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,7 +41,7 @@ python_requires = >=3.6, <3.9
 setup_requires =
     setuptools_scm
 install_requires =
-    attrs
+    attrs >= 20.1.0  # attr.setters.frozen
     jsonschema
     particle
     python-constraint
--- a/src/expertsystem/amplitude/model.py
+++ b/src/expertsystem/amplitude/model.py
@@ -23,7 +23,7 @@ from expertsystem.particle import Particle, ParticleCollection
 
 @attr.s
 class FitParameter:
-    name: str = attr.ib()
+    name: str = attr.ib(on_setattr=attr.setters.frozen)
     value: float = attr.ib()
     fix: bool = attr.ib(default=False)
 
--- a/src/expertsystem/amplitude/model.py
+++ b/src/expertsystem/amplitude/model.py
@@ -28,7 +28,7 @@ class FitParameter:
     fix: bool = attr.ib(default=False)
 
 
-class FitParameters(abc.Mapping):
+class FitParameters(abc.MutableMapping):
     def __init__(
         self, parameters: Optional[Iterable[FitParameter]] = None
     ) -> None:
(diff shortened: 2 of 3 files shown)


redeboer:
@spflueger 
ffd4c36 makes the `FitParameters` mutable, but it doesn't more than doing these kinds of manipulations
https://github.com/ComPWA/expertsystem/blob/ffd4c36e493f613399e4dacba23de7705d1da5ea/tests/unit/amplitude/test_model.py#L65-L73

Ideally, we would want to do something like this:
```python
import expertsystem as es

result = es.generate_transitions(
    initial_state=[("J/psi(1S)", [-1, 1])],
    final_state=["gamma", "pi0", "pi0"],
    allowed_intermediate_particles=["f(0)(980)", "f(0)(1500)"],
    allowed_interaction_types=["strong", "em"],
)
model = es.generate_amplitudes(result)
for particle in result.get_intermediate_particles():
    model.dynamics.set_breit_wigner(particle.name)

# Couple parameters:
model.parameters["Width_f(0)(980)"] = model.parameters["Width_f(0)(1500)"]
assert model.dynamics["Width_f(0)(980)"].pole_imag.name == "Width_f(0)(1500)"
```
but this is not possible with the current setup, because (1) the keys of `FitParameters` need to match `FitParameter.name` and (2) the operation above (even if it would be allowed) would not replace the `FitParameter` that `Dynamics` uses.

Problem (2) might be solved if `FitParameter` were frozen (see [here](https://stackoverflow.com/a/986145)). Problem (1) suggests that `FitParameters` should not a `Collection` (like `ParticleCollection`).

redeboer:
Still, this PR would enable solving #440

redeboer:
Closed by #445

PR #445, opened 2021-01-13
adr: decide on amplitude model structure

redeboer:
- Adds the ADRs back into the Sphinx pages, see [here](https://pwa--445.org.readthedocs.build/projects/expertsystem/en/445/adr.html)


docs/adr/0001-parameters.md
```diff
@@ -0,0 +1,230 @@
...
+
+### SymPY
+
+```{toctree}
+examples/0001/sympy_expression
+```
+
+### Python [`operator`](https://docs.python.org/3/library/operator.html) library
+
+See e.g. [this answer](https://stackoverflow.com/a/7844038) on Stack Overflow
```
spflueger, line 68: I don't get what the `operator` has to do with the parameters

reqs/3.6/requirements-dev.txt
```diff
@@ -160,6 +161,7 @@ sphobjinv==2.0.1
 sqlalchemy==1.3.22
 stdlib-list==0.8.0
 stevedore==3.3.0
+sympy==1.7.1
```
spflueger, line 164: Dp we really want to make the adr add new requirements?

docs/adr/0001-parameters.md
redeboer: It would be needed if you were to create your own 'math language' classes

reqs/3.6/requirements-dev.txt
redeboer: Yes, indeed a bit strange. Currently I only need it for the sympy notebook. But actually the ADR is way too elaborate already.

spflueger: I would say its best if the adrs do not modify the requirements

docs/adr/0001-parameters.md
spflueger: Somehow this adr mixes the whole concept of the `AmplitudeModel`, with the requirements of the fit parameters.

reqs/3.6/requirements-dev.txt
redeboer: See 693d1fd and 1a5dbcc

PR #454, opened 2021-01-20
feat!: formulate amplitude model with SymPy

### LINKED ISSUE #73: Improve recipe file structure
- [x] Add Magnitude and Phase as individual Parameters instead of defining a Type inside Parameter
This will remove some overhead and the code will still be flexible enough. So the content of Magnitude and Phase will just follow Parameter.
- [x] Simplify `DecayProducts`
```yml
DecayParticle:
  Name: J/psi
  Helicity: -1
DecayProducts:
  - Name: f2(1950)
    FinalState: [3, 4]
    Helicity: 0.0
  - Name: gamma
    FinalState: 2
    Helicity: -1
```
- [ ] Refer to object directly instead of `Class` tags
```yml
NormalizedIntensity:
  IncoherentIntensity:
      - CoherentIntensity:
        Component: coherent_J/psi_-1_to_gamma_-1+pi0_0+pi0_0
        Amplitudes:
          CoefficientAmplitude:
            Component: J/psi_-1_to_f2(1270)_0+gamma_-1;f2(1270)_0_to_pi0_0+pi0_0;
            Parameters:
              Magnitude: *par1
              Phase: *par2
            SequentialAmplitude:
                - HelicityDecay:
```

**[Estimated effort: 2h]**

### LINKED ISSUE #124: Add entry point where the user can specify decay dynamics

### LINKED ISSUE #333: Test PreFactor in io.write
In the [`tests/unit/conftest.py`](https://github.com/ComPWA/expertsystem/blob/490c59249e50666b8a7c26a342d9814369aaff48/tests/unit/conftest.py), only J/psi -> gamma pi0 pi0 is generated, which does not contain pre-factors. This resulted in the bug fixed by #332.

### LINKED ISSUE #384: Implement Serializable interface with asdict and fromdict

### LINKED ISSUE #440: Allow inserting custom dynamics

redeboer:
Addressed [ADR-001](https://pwa.readthedocs.io/projects/expertsystem/en/0.6.9/adr/001-amplitude-model.html), see also #445

See also: https://github.com/ComPWA/tensorwaves/pull/222

- [x] Formulate amplitude model in terms of `sympy.Expr`
  - [x] helicity formalism
  - [x] canonical formalism
    - [x] fix `_repr_latex` for `full_expression`
- [x] Formulate class for amplitude model
  - [x] Reconsider `HelicityModel` class (currently results in `model.expression.full_expression`)
  - [x] Reconsider `SuggestedParameterValues` and `ParameterProperties` class
  - [x] Add kinematic variable mapping
  - ~Reconsider naming of `sympy.Symbol`s, e.g.~
    ~`"D[R1 \to 3+4]"` with an additional mapping for resonances, like `{"R1": omega}`~
- [x] Illustrate how to define custom dynamics
- [x] Compute kinematic variables with `numpy`
  (cannot be done with `sympy` due to the need for `einsum`)
  - [x] nested helicity angles
  - [x] invariant masses
  - [x] Handle for data samples: `FourMomenta`, `MomentumPool` etc.

Closes #505
Closes #467
Closes #440
Closes #384 
Closes #333 
Closes #124 
Closes #73


src/expertsystem/amplitude/sympy.py
```diff
@@ -0,0 +1,428 @@
...
+        parent = _HelicityParticle(*graph.get_edge_props(in_edge_ids[0]))
+        children: List[_HelicityParticle] = list()
+        for out_edge_id in out_edge_ids:
+            edge_props = graph.get_edge_props(out_edge_id)
+            children.append(_HelicityParticle(*edge_props))
+
+        wigner_d = Wigner.d(
+            j=sy.Rational(parent.particle.spin, 2),
+            m=sy.Rational(parent.helicity, 2),
+            mp=sy.Rational(children[0].helicity - children[0].helicity, 2),
```
spflueger, line 367: Here the second arguments would have to be dropped, since all the spin values are already at their correct values and would be halved otherwise

src/expertsystem/amplitude/sympy.py
```diff
@@ -0,0 +1,428 @@
...
+            beta=x,
```
spflueger, line 368: The alpha Euler angle is also needed as input (assigned to phi later on)

I also do not get what `x` would be

src/expertsystem/amplitude/sympy.py
```diff
@@ -0,0 +1,428 @@
...
+        self.__create_parameter_couplings(graph_groups)
+        coherent_intensities = [
+            self.__generate_coherent_intensity(graph_group)
+            for graph_group in graph_groups
+        ]
+        if len(coherent_intensities) == 0:
+            raise ValueError("List of coherent intensities cannot be empty")
+        if len(coherent_intensities) == 1:
+            return coherent_intensities[0]
+        return sum(coherent_intensities)
```
spflueger, line 301: Shouldn't this be a sy.sum call? There is another one below.

tests/unit/amplitude/test_sympy_model.py
```diff
@@ -0,0 +1,10 @@
+from expertsystem.amplitude import generate_sympy
+from expertsystem.reaction import Result
+
+
```
spflueger, line 4: We would need a couple of unit tests that validate the correctness of the generated amplitude model. We can try to think of some good tests. Now we also have the possibility to do some mathematical operations like integrals or normal evaluations!

src/expertsystem/amplitude/sympy.py
```diff
@@ -0,0 +1,428 @@
...
+    particles: ParticleCollection = attr.ib(
+        validator=attr.validators.instance_of(ParticleCollection)
+    )
+    intensity_model: sy.Expr = attr.ib(default=None)
+    components: Dict[sy.Function, sy.Function] = attr.ib(default=dict())
+    dynamics: Dict[sy.Function, sy.Function] = attr.ib(default=dict())
+    initial_values: Dict[sy.Symbol, InitialValue] = attr.ib(default=dict())
+
+
+class _SympyHelicityAmplitudeNameGenerator:
```
spflueger, line 59: I did not fully analyze this class, but it seems to be very similar to the existing normal helicity amplitude name generator `_HelicityAmplitudeNameGenerator`.

Is the idea here to just remove the other file once the sympy code has been verified or why was this class/functionality copied?

spflueger (changes requested):
Great work

src/expertsystem/amplitude/sympy.py
redeboer: b0daf78
But `x` indeed still needs to be computed somehow: currently it's just the four-momentum data sample

src/expertsystem/amplitude/sympy.py
redeboer: 890bcc9

src/expertsystem/amplitude/sympy.py
redeboer: Afaik `sy.sum`/`sy.Sum` is for indexed symbols. Here we just have a sum of a list of `sy.Expr`. See also the use of `reduce` for sequential amplitudes elsewhere.

src/expertsystem/amplitude/sympy.py
redeboer: True, this class just started out as a copy of `_HelicityAmplitudeNameGenerator`. The main thing that's different here is that `Particle.latex` is used if available, because then the symbol names render more nicely in the notebooks.

We may have to think about a better naming scheme though... currently the symbol names have become horribly long. So either we provide a mapping from numbered symbols (e.g. `"C_1": "C[J/\\psi(1S) \\to f_{0}(980)_{0} \\gamma_{+1};f_{0}(980) \\to \\pi^{0}_{0} \\pi^{0}_{0}]]'))"`). Or we would have to provide some mapping that tells what those indices mean.

src/expertsystem/amplitude/sympy.py
```diff
@@ -0,0 +1,428 @@
...
+    kinematics: Kinematics = attr.ib(
+        validator=attr.validators.instance_of(Kinematics)
+    )
+    particles: ParticleCollection = attr.ib(
+        validator=attr.validators.instance_of(ParticleCollection)
+    )
+    intensity_model: sy.Expr = attr.ib(default=None)
+    components: Dict[sy.Function, sy.Function] = attr.ib(default=dict())
+    dynamics: Dict[sy.Function, sy.Function] = attr.ib(default=dict())
+    initial_values: Dict[sy.Symbol, InitialValue] = attr.ib(default=dict())
```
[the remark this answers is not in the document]
redeboer: https://github.com/ComPWA/expertsystem/pull/454#discussion_r563622666

src/expertsystem/amplitude/sympy.py
spflueger: A mapping would be most flexible, basically we should always think in terms of that module not in terms of what the user would like to see at the end. So in other words design the dream interface we wish to have for that module and write some adapters or converters for it

tests/unit/amplitude/test_sympy_model.py
spflueger: I already ported some tests here, that perform the angular distribution checks currently done in tensorwaves. The tests also become much more elegant now

redeboer:
@spflueger 
Latest attempt to compute nested helicity angles with `sympy` (this time using `MatrixSymbol`s) can be summarised as this:

<details>
<summary>Code snippet</summary>

```python
# cspell:ignore ndarray
"""Helicity angle computation from ComPWA.

https://github.com/ComPWA/ComPWA/blob/77b491a5dc19f06a24d2d5e4ce535f869b5bb4b2/Physics/HelicityFormalism/HelicityKinematics.cpp#L90-L157
"""

from collections import abc
from typing import Dict, Iterable, Optional, Tuple

import attr
import numpy as np
import sympy as sy
from sympy.matrices.expressions.matadd import MatAdd


def _to_sorted_tuple(instance: Optional[Iterable[int]]) -> Tuple[int, ...]:
    if instance is None:
        return tuple()
    if not isinstance(instance, abc.Iterable):
        raise TypeError(f"Instance {instance} is not iterable")
    if not all(map(lambda i: isinstance(i, int), instance)):
        raise TypeError(f"Not all items in {instance} are {int.__name__}")
    return tuple(sorted(instance))


def _to_sorted_tuple_pair(
    pair: Tuple[Iterable[int], Iterable[int]],
) -> Tuple[Tuple[int, ...], Tuple[int, ...]]:
    if not len(pair) == 2:
        raise ValueError(f"Pair is length {len(pair)}, should be 2")
    return _to_sorted_tuple(pair[0]), _to_sorted_tuple(pair[1])


@attr.s(frozen=True)
class SubSystem:
    final_states: Tuple[Tuple[int, ...], Tuple[int, ...]] = attr.ib(
        converter=_to_sorted_tuple_pair
    )
    recoil_state: Tuple[int, ...] = attr.ib(
        factory=tuple, converter=_to_sorted_tuple
    )
    parent_recoil_state: Tuple[int, ...] = attr.ib(
        default=None, converter=_to_sorted_tuple
    )

    def __attrs_post_init__(self) -> None:
        angle_group, remnant = self.final_states
        common_items = set(angle_group) & set(remnant)
        common_items |= set(angle_group) & set(self.recoil_state)
        common_items |= set(remnant) & set(self.recoil_state)
        if len(common_items) != 0:
            raise ValueError(f"{self} and has common items {common_items}")


def compute_helicity_angles(
    momentum_pool: Dict[int, sy.MatrixSymbol],
    subsystem: SubSystem,
) -> Tuple[sy.Expr, ...]:
    child1_grouping, child2_grouping = subsystem.final_states
    child1 = MatAdd(*[momentum_pool[i] for i in child1_grouping])
    child2 = MatAdd(*[momentum_pool[i] for i in child2_grouping])
    decaying_state = child1 + child2

    # boosting everything into the rest system of the decaying state
    child1 = boost_matrix(decaying_state) * child1

    if subsystem.recoil_state:
        recoil = MatAdd(
            *list(momentum_pool[i] for i in subsystem.recoil_state)
        )
        recoil = boost_matrix(decaying_state) * recoil
        # rotate recoil so that recoil points in the -z direction
        child1 = rotation_matrix_z(-compute_phi(recoil)) * child1
        child1 = rotation_matrix_y(sy.pi - compute_theta(recoil)) * child1

        # in case there is no parent recoil, it is artificially along z
        if subsystem.parent_recoil_state:
            parent_recoil = MatAdd(
                *[momentum_pool[i] for i in subsystem.parent_recoil_state]
            )
            parent_recoil = boost_matrix(decaying_state) * parent_recoil
            parent_recoil = (
                rotation_matrix_z(-compute_phi(recoil)) * parent_recoil
            )
            parent_recoil = (
                rotation_matrix_y(sy.pi - compute_theta(recoil))
                * parent_recoil
            )

            # rotate around the z-axis so that parent recoil lies in xz plane
            child1 = (
                rotation_matrix_z(sy.pi - compute_phi(parent_recoil)) * child1
            )

    return compute_phi(child1), compute_theta(child1)


def boost_matrix(beta: sy.MatrixSymbol) -> sy.Matrix:
    beta = beta / beta[0]
    gamma = compute_gamma(compute_beta(beta))
    b_matrix = gamma * beta[1:, :]
    m_matrix = sy.eye(3) + (gamma - 1) * (b_matrix * b_matrix.T)
    # composition of sub-matrices results in shape (2,2) not (4, 4)
    return sy.Matrix(
        [
            [gamma, *(-b_matrix)],
            [-b_matrix[0], *(m_matrix[0, :])],
            [-b_matrix[1], *(m_matrix[1, :])],
            [-b_matrix[2], *(m_matrix[2, :])],
        ]
    )


def rotation_matrix_y(angle: sy.Expr) -> sy.Matrix:
    return sy.Matrix(
        [
            [1, 0, 0, 0],
            [0, sy.cos(angle), 0, sy.sin(angle)],
            [0, 0, 1, 0],
            [0, -sy.sin(angle), 0, sy.cos(angle)],
        ]
    )


def rotation_matrix_z(angle: sy.Expr) -> sy.Matrix:
    return sy.Matrix(
        [
            [1, 0, 0, 0],
            [0, sy.cos(angle), sy.sin(angle), 0],
            [0, -sy.sin(angle), sy.cos(angle), 0],
            [0, 0, 0, 1],
        ]
    )


def compute_beta(lorentz_vector: sy.MatrixSymbol) -> sy.Expr:
    return p3norm(lorentz_vector) / lorentz_vector[0]


def compute_gamma(beta: sy.Expr) -> sy.Expr:
    return 1 / sy.sqrt(1 - beta ** 2)


def compute_phi(lorentz_vector: sy.MatrixSymbol) -> sy.Expr:
    return sy.atan2(lorentz_vector[2], lorentz_vector[1])


def compute_theta(lorentz_vector: sy.MatrixSymbol) -> sy.Expr:
    r"""Azimuthal angle :math:`\theta` of a Lorentz lorentz_vector."""
    return sy.acos(lorentz_vector[3] / p3norm(lorentz_vector))


def p3norm(lorentz_vector: sy.MatrixSymbol) -> sy.Expr:
    return sy.sqrt(sum(lorentz_vector[1:]))


TEST_DATA = {
    2: np.array(
        [
            (0.514208, -0.184219, 1.23296, 1.35527),
            (0.0727385, -0.0528868, 0.826163, 0.841933),
            (-0.162529, 0.29976, -0.411133, 0.550927),
            (0.0486171, 0.151922, 0.370309, 0.425195),
            (-0.0555915, -0.100214, -0.0597338, 0.186869),
            (0.238921, 0.266712, -1.20442, 1.26375),
            (0.450724, -0.439515, -0.360076, 0.737698),
            (0.552298, 0.440006, 0.644927, 0.965809),
            (-0.248155, -0.158587, -0.229673, 0.397113),
            (1.33491, 0.358535, 0.0457548, 1.38955),
        ]
    ),
    3: np.array(
        [
            (-0.305812, 0.284, -0.630057, 0.755744),
            (0.784483, 0.614347, -0.255334, 1.02861),
            (-0.20767, 0.272796, 0.0990739, 0.356875),
            (0.404557, 0.510467, -0.276426, 0.70757),
            (0.47713, 0.284575, -0.775431, 0.953902),
            (-0.204775, -0.0197981, 0.0799868, 0.220732),
            (0.00590727, 0.709346, -0.190877, 0.734602),
            (0.329157, -0.431973, 0.272873, 0.607787),
            (-0.201436, -0.534829, 0.256253, 0.626325),
            (-0.196357, 0.00211926, -0.33282, 0.386432),
        ]
    ),
    4: np.array(
        [
            (-0.061663, -0.0211864, 0.144596, 0.208274),
            (-0.243319, -0.283044, -0.234866, 0.461193),
            (0.82872, -0.0465425, -0.599834, 1.03294),
            (0.263003, -0.089236, 0.686187, 0.752466),
            (0.656892, -0.107848, 0.309898, 0.746588),
            (0.521569, -0.0448683, 0.43283, 0.692537),
            (-0.517582, -0.676002, -0.0734335, 0.865147),
            (-0.975278, -0.0207817, -0.934467, 1.35759),
            (-0.41665, 0.237646, 0.691269, 0.852141),
            (-0.464203, -0.358114, 0.13307, 0.616162),
        ]
    ),
    5: np.array(
        [
            (-0.146733, -0.0785946, -0.747499, 0.777613),
            (-0.613903, -0.278416, -0.335962, 0.765168),
            (-0.458522, -0.526014, 0.911894, 1.15616),
            (-0.716177, -0.573154, -0.780069, 1.21167),
            (-1.07843, -0.0765127, 0.525267, 1.20954),
            (-0.555715, -0.202046, 0.691605, 0.919879),
            (0.0609506, 0.406171, 0.624387, 0.759452),
            (0.0938229, 0.012748, 0.0166676, 0.165716),
            (0.866241, 0.455769, -0.717849, 1.22132),
            (-0.674348, -0.0025409, 0.153994, 0.704759),
        ]
    ),
}


def test() -> Tuple[np.ndarray, np.ndarray]:
    subsystem = SubSystem(final_states=((3, 4, 5), (2,)))
    state_ids = list(TEST_DATA)
    four_momentum_symbols = [
        sy.MatrixSymbol(f"p_{i}", 4, 1) for i in state_ids
    ]
    four_momentums = [np.array(data) for data in TEST_DATA.values()]
    momentum_pool = dict(zip(state_ids, four_momentum_symbols))
    expr = compute_helicity_angles(momentum_pool, subsystem)
    np_expr = sy.lambdify(four_momentum_symbols, expr, "numpy")
    np_phi, np_theta = np_expr(four_momentums)
    return np_phi, np_theta


if __name__ == "__main__":
    phi, theta = test()
    print(phi)
    print(theta)
```

</details>

I comment it here for documentation purposes, because I think this approach should be abandoned -- for some reason this implementation is so inefficient that the example can't be run.

spflueger:
> @spflueger
> Latest attempt to compute nested helicity angles with [the existing approach in ComPWA](https://github.com/ComPWA/ComPWA/blob/77b491a5dc19f06a24d2d5e4ce535f869b5bb4b2/Physics/HelicityFormalism/HelicityKinematics.cpp#L90-L157) with `sympy` (this time using `MatrixSymbol`s) can be summarised as this:
> 
> Code snippet
> I comment it here for documentation purposes, because I think this approach should be abandoned -- for some reason this implementation is so inefficient that the example can't be run.

Good work! I can check out the code in a bit more detail.

redeboer:
Work on nested helicity angle computation with `sympy` has been moved to
https://github.com/ComPWA/expertsystem/tree/sympy-helicity-angles

docs/conf.py
```diff
@@ -222,7 +237,7 @@
 nb_render_priority["doctest"] = nb_render_priority["html"]
 
 jupyter_execute_notebooks = "off"
-if "EXECUTE_NB" in os.environ:
+if "EXECUTE_NB" in os.environ or "READTHEDOCS" in os.environ:
```
redeboer, line 240: This should be changed back before merging (revert 220625c6)

tests/channels/test_y_to_d0_d0bar_pi0_pi0.py
```diff
@@ -21,8 +21,9 @@ def test_simple(formalism_type, n_solutions, particle_database):
         number_of_threads=1,
     )
     assert len(result.transitions) == n_solutions
-    model = es.generate_amplitudes(result)
-    assert len(model.parameters) == 9
+    model_builder = es.amplitude.get_builder(result)
+    model = model_builder.generate()
+    assert len(model.parameters) == 4
```
spflueger, line 26: why did this change from 9 to 4? Complex amplitude coefficients right?

spflueger (approved):

tests/channels/test_y_to_d0_d0bar_pi0_pi0.py
redeboer: Yes, meson radius for the top decay plus 4 coefficients. These were the parameters previously (produced by running `io.write` in this test for the master branch).
image

PR #459, opened 2021-01-28
refactor: raise violated rules as exception

redeboer:
Violated rules are now raised as an exception. This simplifies the `Result` class: it only contains a formalism type and a list of `StateTransitionGraph`s.

### DIFF
--- a/docs/usage/workflow.ipynb
+++ b/docs/usage/workflow.ipynb
@@ -202,7 +202,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The {meth}`~.StateTransitionManager.find_solutions` method returns a {class}`~.Result` object from which you can extract the {attr}`~.Result.solutions` and any {attr}`~.Result.violated_edge_rules` in case solutions were found. Now, you can use {meth}`~.Result.get_intermediate_particles` to print the names of the intermediate states that the {class}`~.StateTransitionManager` found:"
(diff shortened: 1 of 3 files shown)


redeboer:
> - [ ] Extract an ExecutionInfo class from the QNResult class

@spflueger
I'm trying to find out why 36f19d6 breaks the channel tests

redeboer:
@spflueger 
As a side note, it's also possible to split this PR into:
1. Decorating of classes with `attr.s`. I also tried to decorate `Topology`, but that breaks the tests somehow, so better to tackle that in a separate PR
2. Refactoring of the `Result` class

spflueger:
> > * [ ]  Extract an ExecutionInfo class from the QNResult class
> 
> @spflueger
> I'm trying to find out why [36f19d6](https://github.com/ComPWA/expertsystem/commit/36f19d6e35991c184ae1eedcb2eb9196f1a7f0bf) breaks the channel tests

We already discussed that in private. But for completeness and documentation I comment here as well. So the issue is that invalid `QNResult`s are created (I can't remember the root cause now). Irrelevant of the root cause, the conclusion is, that the design of this Result class has to be changed as a whole. Otherwise the code keeps staying messy somewhere.

spflueger:
> @spflueger
> As a side note, it's also possible to split this PR into:
> 
> 1. Decorating of classes with `attr.s`. I also tried to decorate `Topology`, but that breaks the tests somehow, so better to tackle that in a separate PR
> 2. Refactoring of the `Result` class
> 
> --> #460

I would like to see some plan on how to improve the Result class design first. For example maybe `attr.s` is not needed or things change completely anyways, that it does not help changing the current code into using `attr`.

PR #461, opened 2021-01-28
adr: decide how to implement dynamics

redeboer:
Preview here:
https://pwa--461.org.readthedocs.build/projects/expertsystem/en/461/adr/002.html

### DIFF
--- a/docs/adr/002.md
+++ b/docs/adr/002.md
@@ -38,12 +38,14 @@ They are meant to parametrize a decaying particle.
    it's important that the dynamics can be applied to resonances of some
    **selected** graphs and not generally all graphs in which the resonance
    appears.
-2. The parameters that a dynamics functions requires are registered
+2. The parameters that a dynamics functions requires, are registered
    automatically and linked together.
 3. Where possible, suggested (initial) parameter values are provided as well.
 4. Kinematic variables used in dynamics functions are also linked
    appropriately.
-5. It is easy to define custom dynamics (no boilerplate code)
+5. It is easy to define custom dynamics (no boilerplate code).
+6. It is possible to use and inspect the dynamics expression itself
+   independently from the `expertsystem`.
 
 ## Decision drivers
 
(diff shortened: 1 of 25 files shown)


redeboer:
@spflueger
The three options have been illustrated and analysed here
https://pwa--461.org.readthedocs.build/projects/expertsystem/en/461/adr/002.html#considered-options

I haven't decided on the best one yet and could make the analysis a bit more thorough by ticking the requirements for each of the options, but I think it sketches the main ideas well for now.

docs/adr/002.md
```diff
@@ -0,0 +1,118 @@
...
+`StateTransitionGraph`.
+
+The dynamics expression needs to be formulated in such a way that it satisfies
+[**requirements 2-5**](requirements-adr-002). The following options illustrate
+three different ways of formulating a dynamics expression, taking a
+relativistic Breit-Wigner and a relativistic Breit-Wigner _with form factor_ as
+example.
+
+```{toctree}
+002/composition
```
spflueger, line 101: Im adding my comment here since its a bit hard to find the right places in that notebook. So a couple of comments regarding the cons of the composition solutions:

* Regarding con 1. Why is the expression signature of the parameter dataset form? The expression is just that of a relativistic breit wigner or whatever function is implemented.
* Regarding con 2. I don't understand which argument number checking means here. The function defines what it needs and this has to be mapped with what is available. And at that point the checking can be done. But this as to be done no matter which solution is chosen. I dont really see how this check is automatically guaranteed.
* Regarding con 3. The function code can be included in the builder, just like the other solutions do that as well.
* Regarding con 4. Apart from a `Protocol`, abc / an interface can be defined and enforced.

redeboer: > its a bit hard to find the right places in that notebook

Ok, 9d7adf0 should make this easier. The idea is that the notebooks are only to illustrate implementation and usage, while the main Markdown ADR file is where discussion takes place.

> Regarding con 1. Why is the expression signature of the parameter dataset form? The expression is just that of a relativistic breit wigner or whatever function is implemented.

Could you clarify?

> Regarding con 2. (...) But this as to be done no matter which solution is chosen.

That's true, 3126617 rewrites the Pros and Cons a bit.

> Regarding con 3. The function code can be included in the builder, just like the other solutions do that as well.

I would have to see. I assume that you mean a function is returned from the builder. But that means there is only a builder function that takes an `STG`, so there's no way to use this function independently of the rest of the `expertsystem`.

> Regarding con 4. Apart from a Protocol, abc / an interface can be defined and enforced.

For functions, I think `Protocol` is the only way to enforce the function signature (or `Callable`, but that's just a slimmer version). But I'm not sure if that's a bad thing, see:
https://www.python.org/dev/peps/pep-0544/#rationale-and-goals

spflueger: Was discussed in a meeting. The modified adr will address these points

docs/adr/002.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Considered solutions
+
+To satisfy [**requirement 1**](#solution-requirements), we propose the
+following syntax:
+
+```python
+# model: ModelInfo
+# graph: StateTransitionGraph
+model.set_dynamics(graph, edge_id=1, expression_builder)
```
spflueger, line 88: This `expression_builder` is still a part, I'm quite hesitant about, as this puts a lot of responsibility on the design of how to handle the connection of kinematic variables and parameters. I would simply turn this into an expression, and let some independent code be responsible for the correct construction of that expression (correct meaning creating the correct symbols so that they are connected to other ones in the full model).

spflueger: see https://github.com/ComPWA/expertsystem/pull/461/commits/1241b5cfdbcb68692a34bb5c440ff6a6029fe4cc

spflueger (approved):

PR #469, opened 2021-02-01
adr: reorganize ADR template

redeboer:
The sections of the ADR template has been reorganized as to agree more with upcoming ADRs #458 #461 #467. The idea is that an ADR becomes a bit more like an analysis: state the problem, list requirements of a possible solution, propose some solution, evaluate, and decide which solution to follow.

### DIFF
--- a/docs/adr/001.md
+++ b/docs/adr/001.md
@@ -6,7 +6,7 @@
 - Status: **accepted**
 - Deciders: @redeboer @spflueger
 
-## Context and Problem Statement
+## Context and problem statement
 
 From the perspective of a PWA fitter package, the responsibility of the
 `expertsystem` is to construct an `AmplitudeModel` that serves as blueprint for
@@ -45,7 +45,7 @@ Requirements for the `AmplitudeModel`:
    for the calculation of fit fractions, or plotting individual parts of the
    intensity.
 
-## Decision Drivers
+## Decision drivers
 
 - [#382](https://github.com/ComPWA/expertsystem/issues/382): Coupling
   parameters in the `AmplitudeModel` is difficult (has to be done through the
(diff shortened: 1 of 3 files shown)


docs/adr/001.md
```diff
@@ -45,7 +45,7 @@ Requirements for the `AmplitudeModel`:
    for the calculation of fit fractions, or plotting individual parts of the
    intensity.
 
-## Decision Drivers
+## Decision drivers
```
redeboer, line 48: Note the style change: titles are not capitalized anymore to avoid problems in the sidebar:
image

This could be read as:
- Contex and Problem
- Statement
- ...

but was actually
- Contex and Problem Statement
- ...

docs/adr/template.md
```diff
@@ -9,39 +9,90 @@
...
+ADR:
+
+- [#000](https://github.com/ComPWA/expertsystem/issues/000)
+- [#000](https://github.com/ComPWA/expertsystem/issues/000)
+-->
+
+### Solution requirements
+
+1. <!-- Requirement 1 -->
+2. <!-- Requirement 2 -->
```
spflueger, line 39: I like the word decision drivers, as its maybe not always possible to define strict requirements. And decision drivers is generally just saying which aspects influence your decision making.

Maybe we can do something like this:

```md
## Decision drivers

### Solution requirements

1. <!-- Requirement 1 -->
￼2. <!-- Requirement 2 -->
```

redeboer: 371aca1

docs/adr/template.md
```diff
@@ -9,39 +9,92 @@
...
+- **Positive**:
+  1.
+  2.
+- **Negative**:
+  1.
+  2.
+-->
 
-#### Cons
+### Requirement evaluation
```
spflueger, line 85: I would make this optional

redeboer: 5429fc5

spflueger (approved):

PR #470, opened 2021-02-03
feat: freeze Topology class

redeboer:
The `StateTransitionGraph` is a mapping of properties onto an underlying topology structure. It therefore behaves like a collection of `dict`s, whether the keys are the edge and node IDs of the topology. Currently, the underlying `Topology` is, however, mutable and therefore does not serve well as a collection to map properties onto.

This results in subtle bugs. For instance, the `StateTransitionGraph` falsely protects its underlying `Topology` by dundering it and not exposing it as a `property`. I say falsely, because the `Topology.edges` are still exposed (example from the `stable` branch):

https://github.com/ComPWA/expertsystem/blob/1fb82ffabb1d384454b3790f7689e215ade12f73/src/expertsystem/reaction/topology.py#L369-L371

This allows one to modify the underlying topology as follows:

```python
# graph: StateTransitionGraph
graph.edges[666] = "gotcha!"
```

Another problem is that references to the same topology are being carried around by `ProblemSet` and `QNProblemSet`. There is no assurance that they change 'their' `topology` somewhere else in the code.

**Additional motivation**
Eventually, I would like to introduce a new class in addition to the `StateTransitionGraph`. This class (`StateTransition`) is an **immutable and hashable** mapping of `Particle` instances with a `float` spin projection to the edges of a topology. An overarching `StateTransitionCollection` (a `set` of `StateTransition`s) then provides a complete description of the allowed state transitions between some initial and final states.

The key difference between `StateTransitionGraph` and `StateTransition` is that the latter is immutable (and can therefore be made hashable). The `STG` is then only used within the `reaction` module (particularly `solving`), while `StateTransitionCollection` is the product that the `reaction` submodule produces. In that sense, we generally get:

- `expertsystem.particle` produces a `ParticleCollection`
- `expertsystem.reaction` produces a `StateTransitionCollection`
- `expertsystem.reaction` produces a `sympy.Expr` (see [ADR-001](https://pwa.readthedocs.io/projects/expertsystem/en/latest/adr/001.html))

At any rate, this is yet another class that maps onto an underlying topology. So I do not want to have this risk of carrying around a reference to some mutable topology as well as having to falsely hide it and then needing to yet again provide getter methods like `get_final_state_edge_ids`.

More thoughts at https://github.com/ComPWA/expertsystem/pull/470#issuecomment-772589710

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -161,6 +161,8 @@ nitpick_ignore = [
     ("py:class", "a set-like object providing a view on D's keys"),
     ("py:class", "an object providing a view on D's values"),
     ("py:class", "typing_extensions.Protocol"),
+    ("py:obj", "expertsystem.reaction.topology._K"),
+    ("py:obj", "expertsystem.reaction.topology._V"),
 ]
 
 # Intersphinx settings
--- a/src/expertsystem/amplitude/canonical_decay.py
+++ b/src/expertsystem/amplitude/canonical_decay.py
@@ -82,10 +82,11 @@ def _clebsch_gordan_decorator(
             raise ValueError(
                 f"{spin.__class__.__name__} is not of type {Spin.__name__}"
             )
+        topology = graph.topology
+        in_edge_ids = topology.get_edge_ids_ingoing_to_node(node_id)
+        out_edge_ids = topology.get_edge_ids_outgoing_from_node(node_id)
 
-        in_edge_ids = graph.get_edge_ids_ingoing_to_node(node_id)
         in_edge_id = next(iter(in_edge_ids))
-
         particle, spin_projection = graph.get_edge_props(in_edge_id)
         parent_spin = Spin(
             particle.spin,
@@ -93,8 +94,7 @@ def _clebsch_gordan_decorator(
         )
 
         daughter_spins: List[Spin] = []
-
-        for out_edge_id in graph.get_edge_ids_outgoing_from_node(node_id):
+        for out_edge_id in out_edge_ids:
             particle, spin_projection = graph.get_edge_props(out_edge_id)
             daughter_spin = Spin(
                 particle.spin,
--- a/src/expertsystem/amplitude/canonical_decay.py
+++ b/src/expertsystem/amplitude/canonical_decay.py
@@ -40,7 +40,7 @@ class _CanonicalAmplitudeNameGenerator(_HelicityAmplitudeNameGenerator):
         if isinstance(node_id, int):
             node_ids = frozenset({node_id})
         else:
-            node_ids = graph.nodes
+            node_ids = graph.topology.nodes
         for node in node_ids:
             name += (
                 super().generate_unique_amplitude_name(graph, node)[:-1]
--- a/src/expertsystem/amplitude/helicity_decay.py
+++ b/src/expertsystem/amplitude/helicity_decay.py
@@ -327,8 +327,9 @@ class _HelicityAmplitudeNameGenerator:
     def _retrieve_helicity_info(
         graph: StateTransitionGraph[ParticleWithSpin], node_id: int
     ) -> Tuple[List[Tuple[str, float]], List[Tuple[str, float]]]:
-        in_edges = graph.get_edge_ids_ingoing_to_node(node_id)
-        out_edges = graph.get_edge_ids_outgoing_from_node(node_id)
+        topology = graph.topology
+        in_edges = topology.get_edge_ids_ingoing_to_node(node_id)
+        out_edges = topology.get_edge_ids_outgoing_from_node(node_id)
 
         in_names_hel_list = _get_name_hel_list(graph, in_edges)
         out_names_hel_list = _get_name_hel_list(graph, out_edges)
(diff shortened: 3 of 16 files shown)


redeboer:
@spflueger
For review purposes, I could split up this PR. The [commits](https://github.com/ComPWA/expertsystem/pull/470/commits) have been kept clean in that four separate things happen:
1. fix or improve type hints (original motivation was to switch from `Sequence` to `Iterable` because we can now have a `set` of topologies)
2. freeze `Topology`
3. improve the topology builder interface and have it return a `Set[Topology]`

spflueger:
> The `StateTransitionGraph` is a mapping of properties onto an underlying topology structure. It therefore behaves like a collection of `dict`s, whether the keys are the edge and node IDs of the topology. Currently, the underlying `Topology` is, however, mutable and therefore does not serve well as a collection to map properties onto.
> 

I don't understand why a topology has to be immutable to make another object immutable which has a topology attribute. So why is a private attribute not enough?

> Note: I would like to introduce a new class in addition to the `StateTransitionGraph`. This class (`StateTransition`) is an immutable mapping of `Particle` instances with a `float` spin projection to the edges of a topology. An overarching `StateTransitionCollection` (a `set` of `StateTransition`s) then provides a complete description of the allowed state transitions between some initial and final states.
> 
> The key difference between `StateTransitionGraph` and `StateTransition` is that the latter is immutable. The `STG` is then only used within the `reaction` module (particularly `solving`), while `StateTransitionCollection` is the product that the `reaction` submodule produces. In that sense, we generally get:
> 
So StateTransitionGraph is almost immutable except the `swap_edges` method. We could simply turn this function into a similar one as evolve that does not modify the current instance but returns a new one. Why is that not enough?

> * `expertsystem.particle` produces a `ParticleCollection`
> * `expertsystem.reaction` produces a `StateTransitionCollection`
> * `expertsystem.reaction` produces a `sympy.Expr` (see [ADR-001](https://pwa.readthedocs.io/projects/expertsystem/en/latest/adr/001.html))

redeboer:
https://github.com/ComPWA/expertsystem/pull/470#issuecomment-772387455

Ok, I narrowed down the [issue description](https://github.com/ComPWA/expertsystem/pull/470#issue-566672266) a bit in an attempt to formulate what I was trying to achieve intuitively. Some additional thoughts (not perfectly worked out):

**(1) Hashable: solutions are unique**
The `reaction` module of the `expertsystem` generally tries to find a set of solutions for some problem set. This ranges from finding a collection of isobar topologies for a number of incoming and outgoing lines to finding a collection of allowed state transitions given some initial and final states plus constraints. Given this language, I would say that solutions have to be hashable.

This offers some additional advantages. For instance, when investigating the allowed state transitions for 1-to-4 body decays, you may want to select only state transitions with the same topology. This works best if that topology is hashable.

**(2) Read-only versus immutable**
I think we should use `attrs` as much as possible, as that not only avoids boilerplate code, but also enforces the developer to write safer Python code. For instance, `attrs` [believes](https://www.attrs.org/en/20.3.0/init.html#private-attributes) that Python does not have things like private methods or members. (Note: public, private or protected are not even listed in [Python's glossary](https://docs.python.org/3/glossary.html), while immutable etc. is.)

Therefore, `attrs` achieves immutability by preventing data members from being set, not with name mangling (dunders):

```python
from typing import Dict
import attr

@attr.s(frozen=True)
class FrozenClass:
    mapping: Dict[int, str] = attr.ib(factory=dict)

frozen_class = FrozenClass({1: "one", 2: "two"})
try:
    frozen_class.mapping = {1: "three"}
except attr.exceptions.FrozenInstanceError:
    print("See, didn't work")
```

Note that it is still possible to modify members if they are mutable (just as if `FrozenClass` would have been an immutable built-in container like a `tuple`):
```
frozen_class.mapping[1] = "three"  # allowed
print(frozen_class)
```

If you want some member of a data container not to be modified, it means that _the data member itself should be immutable_. This is safer than using dunders, because there is no way tell whether the dunder member is meant to be a read-only `property` to the outside or whether, in addition, this member is not supposed to be changed by the class itself.

This introduces more subtle bugs. What happens for instance to the `topology` attribute (passed on as a reference!) when you pass it on to `QNProblemSet`?

https://github.com/ComPWA/expertsystem/blob/0ca61135a784ef3555c2844521d6d5f334e83d15/src/expertsystem/reaction/solving.py#L556-L574

And can we be sure that the `topology` isn't modified if you give it to this function?
https://github.com/ComPWA/expertsystem/blob/0ca61135a784ef3555c2844521d6d5f334e83d15/src/expertsystem/reaction/combinatorics.py#L160-L163

It seems far-fetched that one would make mistakes in the above examples, but the fact that `Topology` is mutable already caused bugs in the past:
https://github.com/ComPWA/expertsystem/blob/1fb82ffabb1d384454b3790f7689e215ade12f73/src/expertsystem/reaction/topology.py#L320-L321

**(3) Property mappings**

Currently, three classes carry around a `topology` (`ProblemSet`, `QNProblemSet`, and `StateTransitionGraph`). This might be extended with an additional `StateTransition` class. All these classes provide some mapping of properties onto edges and nodes of the underlying topology. For the reasons mentions above, I think the responsibility of immutability should therefore really lie with `Topology` itself, not with the classes (or functions) that work with instances of it.

PR #482, opened 2021-02-05
test: check identical particle combinatorics

redeboer:
Adds a test for `perform_external_edge_identical_particle_combinatorics`.


redeboer:
@spflueger 
Currently only testing Y --> D0 D0bar pi0 pi0. This is probably not as interesting though. This is, however, the only 4-body decay channel that can be generated quickly (because spin projection 0).

tests/unit/reaction/test_combinatorics.py
```diff
@@ -206,3 +208,23 @@ def test_generate_permutations(
     assert spin_permutations[1][2][1] == +1
     assert spin_permutations[2][0][1] == +1
     assert spin_permutations[3][0][1] == +1
+
+
+def test_perform_external_edge_identical_particle_combinatorics(
+    y_to_d0_d0bar_pi0_pi0,
```
spflueger, line 214: You dont really need a full solution graph here. So creating some artificial dummy graph would do it.

tests/unit/reaction/test_combinatorics.py
```diff
@@ -206,3 +208,23 @@ def test_generate_permutations(
...
+):
+    result: Result = y_to_d0_d0bar_pi0_pi0("helicity")
+    some_graph = next(iter(result.transitions))
+    outgoing_edge_ids = some_graph.topology.outgoing_edge_ids
+    final_states = {
+        i: some_graph.get_edge_props(i)[0].name for i in outgoing_edge_ids
+    }
+    assert final_states == {3: "D0", 4: "pi0", 5: "D~0", 6: "pi0"}
```
spflueger, line 222: Not bad as a first test. Things would get more interesting if there would be final states with two pairs of identical particles  (such as pi-pi-pi+pi+) or 3 or 4 identical particles (such as 3/4 x pi0).

redeboer:
@spflueger 
I have the impression dumplicate solutions are not filtered out correctly. With #487 and the `groupings` notebook introduced by 5444062, I generated the following transitions. Judging from the labels, there are duplicates.

- [particle_graphs.pdf](https://github.com/ComPWA/expertsystem/files/5940072/particle_graphs.pdf)
- [transitions.pdf](https://github.com/ComPWA/expertsystem/files/5940073/transitions.pdf)

spflueger:
> @spflueger
> I have the impression duplicate solutions are not filtered out correctly. With #487 and the `groupings` notebook introduced by [b049d3e](https://github.com/ComPWA/expertsystem/commit/b049d3ef81a4a7104d83b345b524b19d6f46a81c), I generated the following transitions. Judging from the labels, there are duplicates.
> 
> * [particle_graphs.pdf](https://github.com/ComPWA/expertsystem/files/5940072/particle_graphs.pdf)
> * [transitions.pdf](https://github.com/ComPWA/expertsystem/files/5940073/transitions.pdf)

Does this part already include the final state identical particle symmetrization? Basically the final state particle unique ids would have to be visible in these pictures as well to say if they are identical

docs/usage/groupings.ipynb
```diff
@@ -0,0 +1,131 @@
...
+    "    [InteractionTypes.Strong, InteractionTypes.EM]\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "covered-myrtle",
+   "metadata": {},
+   "source": [
+    "We are expecting the resonance to be between the `\\pi^+\\pi^-` pairs. This can be indicated with {meth}`.StateTransitionManager.add_final_state_grouping`:"
```
spflueger, line 62: The `\\pi^+\\pi^-` does not render as latex in the docu

tests/channels/test_y_to_d0_d0bar_pi0_pi0.py
```diff
@@ -33,19 +33,8 @@ def test_simple(formalism_type, n_solutions, particle_database):
...
-        allowed_intermediate_particles=["D*"],
-        formalism_type=formalism_type,
-        number_of_threads=1,
-    )
-    stm.set_allowed_interaction_types([InteractionTypes.Strong])
-    stm.add_final_state_grouping([["D0", "pi0"], ["D~0", "pi0"]])
-    problem_sets = stm.create_problem_sets()
-    result = stm.find_solutions(problem_sets)
+def test_full(y_to_d0_d0bar_pi0_pi0, formalism_type, n_solutions):
+    result: Result = y_to_d0_d0bar_pi0_pi0(formalism_type)
```
spflueger, line 37: This can be changed back again?

redeboer:
> Does this part already include the final state identical particle symmetrization?

It's standard output of the `STM`, which [has `identical_particle_symmetrization` by default](https://github.com/ComPWA/expertsystem/blob/0344f84868bfdb90bcfc76a32a8f4cd565a89420/src/expertsystem/reaction/default_settings.py#L210-L219), so I guess yes.

> Basically the final state particle unique ids would have to be visible in these pictures as well to say if they are identical

Does the final state ID matter in the output of the `reaction` module? I still would like to move towards a system where the `reaction` module tells which decay topologies are allowed (filtering out isomorphic ones). Then the `amplitude` model has the responsibility to put back final state combinatorics if needed.

tests/unit/reaction/test_combinatorics.py
```diff
@@ -206,3 +213,49 @@ def test_generate_permutations(
...
+    )
+    get_originating_node = lambda g: tuple(  # noqa: E731
+        g.topology.edges[i].originating_node_id for i in final_state_edge_ids
+    )
+    for g in graphs:  # pylint: disable=invalid-name
+        assert get_final_state(g) == ("pi-", "pi+", "pi-", "pi+")
+    assert get_originating_node(graphs[0]) == (1, 1, 2, 2)
+    assert get_originating_node(graphs[1]) == (1, 2, 2, 1)
+    assert get_originating_node(graphs[2]) == (2, 1, 1, 2)
+    assert get_originating_node(graphs[3]) == (2, 2, 1, 1)
```
spflueger, line 261: This is quite interesting actually. In this example (using this topology) there are actually only 3 valid final state combinations! So we have to clearly specify what this code is supposed to do. So is this function supposed to generate all combinations, or only the physically meaningful? I have to dig a bit deeper here as well

redeboer: Indeed, seems it comes down to what we want the `reaction` module to return as a whole. What is the meaning of 'solutions'?

docs/usage/groupings.ipynb
redeboer: Thanks, well spotted! --> feeda13

tests/channels/test_y_to_d0_d0bar_pi0_pi0.py
redeboer: Yep --> 06f7469

tests/unit/reaction/test_combinatorics.py
spflueger: Solution currently means one completely specific amplitude transition. So changing the slightest bit of info which is physically not the identical, would result in a different solution.

So to me `perform_external_edge_identical_particle_combinatorics` suggests it will only create the physically relevant combinations. Also the code doc string explicitly states that only identical particles originating from different nodes can be swapped. So only the physically relevant combinations is def the responsibility of this function.

What I did not realize just now is that it also matters what kind of state is assigned to the intermediate edges... So actually this result would actually be correct. And we can do this combinatorics generally, even if the two intermediate state would be identical in this example. Then theoretically swapping the two final state groups completely (11, 22) vs (22, 11) are actually the same solution, but with the correct normalization (1/2 on amp level), `sympy.simplify` can combine these two again.

So all in all, i would say the code is fine and doing what it should.

spflueger:
> > Does this part already include the final state identical particle symmetrization?
> 
> It's standard output of the `STM`, which [has `identical_particle_symmetrization` by default](https://github.com/ComPWA/expertsystem/blob/0344f84868bfdb90bcfc76a32a8f4cd565a89420/src/expertsystem/reaction/default_settings.py#L210-L219), so I guess yes.
> 
Ok bad choice of words from my side. I meant this identical particle combinatorics. Afaik this is done in the amplitude building, so i should not be included here?

> > Basically the final state particle unique ids would have to be visible in these pictures as well to say if they are identical
> 
> Does the final state ID matter in the output of the `reaction` module? I still would like to move towards a system where the `reaction` module tells which decay topologies are allowed (filtering out isomorphic ones). Then the `amplitude` model has the responsibility to put back final state combinatorics if needed.

I'm trying to understand what is plotted here. From the picture there are duplicate solutions, which would be a bug. So we should fix this asap. But if those plots are generated on the amplitude level, then there can be final state symmetrization, and a factor of 2 identical plots in this form would be expected (as the final state mappings are not the same).

redeboer:
https://github.com/ComPWA/expertsystem/pull/482#issuecomment-775032444
Small summary again from internal discussion:
- Seems indeed that there is a combinatorics bug --> #489
- The cause is probably `match_external_edges`. It has to take intermediate edges into account and/or it has to be executed on a different place in `StateTransitionManager.find_solutions`.
- `perform_external_edge_identical_particle_combinatorics` does behave as expected in the test. I would therefore say this PR can be merged (I kicked out the new notebook, that has to be fixed through #489 first)

PR #484, opened 2021-02-05
feat: add option to change width factor

### LINKED ISSUE #355: Allow user to modify width in mass conservation rule
I noticed that the expertsystem uses a 5*width as a boundary condition for the resonances that are outside of the phase space. Here is the MassConservation defined for a node:

https://github.com/ComPWA/expertsystem/blob/74cb82821d6bcea9a84f6a70216be0abca268388/expertsystem/reaction/_default_settings.py#L159

The questions are:

1.  Is 5* width reasonable

2.  Is there a (easy) way to change it? I remember that there was a way in ComPWA but I can't recall. If there is a way it should be mentioned in the tutorials and if not that it should be implemented.

redeboer:
Closes #355

A better solution would come through #169, but for now this is a necessary feature that only adds to the interfaces.


spflueger (approved):
GJ

PR #486, opened 2021-02-07
test: check create_default_interaction_settings

redeboer:


spflueger:
What is the idea or purpose of this PR?

redeboer:
The domains would follow the domains that are allowed by the `ParticleCollection` that you use. So if you use a small subset it would become smaller than the one than the default, and if you use some exotic particles that would fall outside the domains, the domains increase accordingly.

spflueger:
> The domains would follow the domains that are allowed by the `ParticleCollection` that you use. So if you use a small subset it would become smaller than the one than the default, and if you use some exotic particles that would fall outside the domains, the domains increase accordingly.

Ok that part I got. But why? So the only reason is that its speeding up the calculation a bit. So the ideal situation would be to work as long as possible on the quantum numbers themselves and introduce the particle only at a later point. The reason is to get a full picture of what is physically allowed and what not. Using the particles that are currently known to narrow down the qn domains would be quite dangerous in my opinion.

redeboer:
https://github.com/ComPWA/expertsystem/pull/486#issuecomment-775047904
Ok, just to summary:
- This PR came from the idea that the `reaction` module gets `Particle`s as input and returns a `StateTransitionGraph` with `Particle`s as output. The ideal is, however, that the `expertsystem.reaction` module works with sets of quantum numbers only, irrespective of whether those quantum number sets are known particles (as listed in the PDG).
- The changes that this PR makes to the `default_settings` module will therefore be reverted, as they are not in line with the aims of the `expertsystem`.
- What can still be done: rewrite the `test_default_settings` that this PR introduces to test also cases like `nbody_topology=True`.

redeboer:
> What can still be done: rewrite the test_default_settings that this PR introduces to test also cases like nbody_topology=True.

Done :heavy_check_mark:

PR #487, opened 2021-02-07
feat: render node properties in dot language

redeboer:
This allows one to draw images like these:
<details>
<summary>Source code</summary>

```python
import graphviz
import expertsystem as es
from expertsystem.reaction import InteractionTypes, StateTransitionManager


stm = StateTransitionManager(
    initial_state=[("J/psi(1S)", [+1])],
    final_state=["pi-", "pi+", "pi-", "pi+"],
    allowed_intermediate_particles=["f(0)(980)", "rho(770)0"],
    formalism_type="canonical",
)
stm.set_allowed_interaction_types([InteractionTypes.Strong, InteractionTypes.EM])
stm.add_final_state_grouping([["pi-", "pi+"], ["pi-", "pi+"]])
problem_sets = stm.create_problem_sets()
result = stm.find_solutions(problem_sets)

dot = es.io.convert_to_dot(result.get_particle_graphs())
graphviz.Source(dot).view()
```

</details>

image

### DIFF
--- a/src/expertsystem/amplitude/helicity_decay.py
+++ b/src/expertsystem/amplitude/helicity_decay.py
@@ -261,8 +261,7 @@ class _HelicityAmplitudeNameGenerator:
                 priority_partner_coefficient_suffix,
             ) = self._generate_amplitude_coefficient_couple(graph, node_id)
 
-            node_props = graph.get_node_props(node_id)
-            if node_props is not None and node_props.parity_prefactor is None:
+            if graph.get_node_props(node_id).parity_prefactor is None:
                 continue
 
             if (
--- a/src/expertsystem/amplitude/helicity_decay.py
+++ b/src/expertsystem/amplitude/helicity_decay.py
@@ -261,7 +261,8 @@ class _HelicityAmplitudeNameGenerator:
                 priority_partner_coefficient_suffix,
             ) = self._generate_amplitude_coefficient_couple(graph, node_id)
 
-            if graph.get_node_props(node_id).parity_prefactor is None:
+            node_props = graph.get_node_props(node_id)
+            if node_props is not None and node_props.parity_prefactor is None:
                 continue
 
             if (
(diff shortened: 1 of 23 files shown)


src/expertsystem/amplitude/helicity_decay.py
```diff
@@ -261,7 +261,8 @@ def register_amplitude_coefficient_name(
                 priority_partner_coefficient_suffix,
             ) = self._generate_amplitude_coefficient_couple(graph, node_id)
 
-            if graph.get_node_props(node_id).parity_prefactor is None:
+            node_props = graph.get_node_props(node_id)
+            if node_props is not None and node_props.parity_prefactor is None:
```
spflueger, line 265: Will the new code also crash if there are no node props defined or just build something which is incorrect?

src/expertsystem/reaction/topology.py
```diff
@@ -571,8 +571,8 @@ def __post_init__(self) -> None:
         _assert_over_defined(self.topology.nodes, self.__node_props)
         _assert_over_defined(self.topology.edges, self.__edge_props)
 
-    def get_node_props(self, node_id: int) -> InteractionProperties:
-        return self.__node_props[node_id]
+    def get_node_props(self, node_id: int) -> Optional[InteractionProperties]:
+        return self.__node_props.get(node_id)
```
spflueger, line 575: Why did the `node_props` become optional now? The reason why i did not make this optional, is because I require those to exist. So that all of the other code working with the graphs can assume they exist

redeboer: If they should not be optional, we should check that all nodes have properties assigned. I remember that if I didn't use `dict.get` when running the notebooks with this new visualization procedure, the program would crash because of missing keys. But I would have to recheck; these commits have been extracted from a branch that had the work on the other PRs combined.

redeboer: > (...) the program would crash because of missing keys.

I checked, apparently the cause for that was something else. 05ef4cb reverts the `Optional` change.

src/expertsystem/amplitude/helicity_decay.py
redeboer: 05ef4cb5f reverts this

spflueger (approved):

PR #497, opened 2021-02-17
fix: remove cell ids from notebooks

redeboer:
Somehow if a Jupyter notebook contains
```
  "nbformat_minor": 5
```
in its [top-level structure](https://ipython.org/ipython-doc/dev/notebook/nbformat.html#top-level-structure), instead of
```
  "nbformat_minor": 4
```
both `nbstripout` and Jupyter lab add randomly generated cell IDs to each cell. This is super bothersome for diffs and merge conflicts. Worst part is that the IDs are random which causes a lot of unnecessary file changes.

This PR removes the `"id"` items and lowers the `nbformat_minor`. With the current dev set-up (pinned versions of `jupyterlab` etc) new notebooks will again get `"nbformat_minor": 5`, but I'm not sure how to work around this. Just something to keep an eye on now.

### DIFF
--- a/docs/adr/001/operators.ipynb
+++ b/docs/adr/001/operators.ipynb
@@ -2,7 +2,6 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "thick-yesterday",
    "metadata": {},
    "source": [
     "# Python `operator` library"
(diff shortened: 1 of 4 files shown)


redeboer:
Note: I initially suspected [`"hide-cell"` tags](https://myst-nb.readthedocs.io/en/latest/use/hiding.html) or [`"keep_output"`](https://github.com/kynan/nbstripout#keeping-some-output), but it's really `"nbformat_minor": 5` that causes these IDs.

PR #513, opened 2021-03-12
refactor!: remove io module

### LINKED ISSUE #383: Remove io module

redeboer:
Closes #383

Since #454, the `io` module has shrunk considerable. This PR shrinks if further by moving the `load_pdg` function to the `particle` module. Some additional changes:

- `io.convert_to_dot` has been renamed to `io.asdot`, in accordance with `asdict`/`fromdict`. #515 could be a follow-up to this by introducing `io.fromdot`.
- Redundant files like the validation scheme for the amplitude model and `expected_recipe.yml` in the tests have been removed.
- The `schemas` folder has been disolved into the `particle` module.
- Hidden sub-modules of `io._dict` has been merged into one.


spflueger:
I'm fine getting rid of that `io` module, but we should pick one side and stick with it. For me it seems like a rather subjective decision. Heres an overview:

* **io module**: io logic is grouped in one place. Individual io logic can still be somewhere else, its just a ui facade.
   * **pro**:
       * potentially easier to comprehend (this is highly subjective)
       * keeping io import level high (just import expertsystem.io)
   * **cons**:
      * extra code
      * (subjective pro is not really pro if you dont like it)
* **w/o io module**: io interface is directly in the module.
   * **pro**:
       * potentially easier to comprehend (this is highly subjective)
       * no extra code needed
       * io functionality comes with import of the module itself
   * **cons**:
       * io import level is mixed (but is grouped with the module, so usually no extra import is needed)

redeboer:
I decided to leave the `io` module there and have this PR only clean up things a bit. The responsibility of the `io` module is clear. If that module would be dissolved into other modules, they clogged with more code.

PR #519, opened 2021-03-15
feat: export and import transitions to JSON/YAML

redeboer:
Allow to serialize `Result`, `StateTransitionGraph[ParticleWithSpin]`, and `Topology` to and from `dict` with `io.asdict`/`io.fromdict` so that they can be written to disk with `io.write`. This is useful if it takes a long time to generate transitions.

Some changes were required:
- `load_default_particles` had to be moved to the top import, so that the `reaction` module is not dependent on the `io` module (the type hint for `Result` in the `io` introduces a circular dependency).
- Some converters were added to the `attr`-decorated classes that are being serialized, so that they still work after serializing to and from JSON (which has `str` keys only).
- Added a 'strict' `__eq__` method for `StateTransitionGraph` for testing.

### DIFF
--- a/src/expertsystem/io/_dict.py
+++ b/src/expertsystem/io/_dict.py
@@ -12,6 +12,7 @@ from expertsystem.particle import Parity, Particle, ParticleCollection, Spin
 from expertsystem.reaction import (
     InteractionProperties,
     ParticleWithSpin,
+    Result,
     StateTransitionGraph,
     Topology,
 )
@@ -29,6 +30,15 @@ def from_particle(particle: Particle) -> dict:
         value_serializer=__value_serializer,
         filter=lambda attr, value: attr.default != value,
     )
+
+
+def from_result(result: Result) -> dict:
+    output: Dict[str, Any] = {
+        "transitions": [from_stg(graph) for graph in result.transitions],
+    }
+    if result.formalism_type is not None:
+        output["formalism_type"] = result.formalism_type
+    return output
 
 
 def from_stg(graph: StateTransitionGraph[ParticleWithSpin]) -> dict:
@@ -102,6 +112,17 @@ def build_particle(definition: dict) -> Particle:
         if parity_def is not None:
             definition[parity] = Parity(**parity_def)
     return Particle(**definition)
+
+
+def build_result(definition: dict) -> Result:
+    formalism_type = definition.get("formalism_type")
+    transitions = [
+        build_stg(graph_def) for graph_def in definition["transitions"]
+    ]
+    return Result(
+        transitions=transitions,
+        formalism_type=formalism_type,
+    )
 
 
 def build_stg(definition: dict) -> StateTransitionGraph[ParticleWithSpin]:
--- a/src/expertsystem/io/_dict.py
+++ b/src/expertsystem/io/_dict.py
@@ -9,7 +9,11 @@ import attr
 import jsonschema
 
 from expertsystem.particle import Parity, Particle, ParticleCollection, Spin
-from expertsystem.reaction.topology import Edge, Topology
+from expertsystem.reaction.quantum_numbers import (
+    InteractionProperties,
+    ParticleWithSpin,
+)
+from expertsystem.reaction.topology import Edge, StateTransitionGraph, Topology
 
 
 def from_particle_collection(particles: ParticleCollection) -> dict:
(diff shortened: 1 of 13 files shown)


src/expertsystem/io/_dict.py
```diff
@@ -24,12 +32,55 @@ def from_particle(particle: Particle) -> dict:
     )
 
 
+def from_result(result: Result) -> dict:
```
spflueger, line 35: Feels like this should be part of the specific class. What was the reasoning not to do that?

src/expertsystem/io/_dict.py
```diff
@@ -63,6 +114,49 @@ def build_particle(definition: dict) -> Particle:
     return Particle(**definition)
 
 
+def build_result(definition: dict) -> Result:
```
spflueger, line 117: Same as above.

src/expertsystem/reaction/__init__.py
```diff
@@ -483,7 +481,7 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-branches
             )
 
         if reload_pdg or len(self.__particles) == 0:
-            self.__particles = load_default_particles()
+            self.__particles = load_pdg()
```
spflueger, line 484: Isn't the full default particle list needed here? So pdg + the additional particle definitions?

redeboer: Can't import from the top module in the `reaction` module:

> load_default_particles had to be moved to the top import, so that the reaction module is not dependent on the io module (the type hint for Result in the io introduce a circular dependency).

src/expertsystem/reaction/topology.py
```diff
@@ -78,12 +78,22 @@ def values(self) -> ValuesView[_V]:
...
+
+
 @attr.s(frozen=True)
 class Edge:
     """Struct-like definition of an edge, used in `Topology`."""
 
-    originating_node_id: Optional[int] = attr.ib(default=None)
-    ending_node_id: Optional[int] = attr.ib(default=None)
+    originating_node_id: Optional[int] = attr.ib(
+        default=None, converter=_to_optional_int
```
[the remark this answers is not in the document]
redeboer: Yeah, it's for JSON, because that sometimes casts to `str`.

src/expertsystem/io/_dict.py
redeboer: Just the way it was already implemented for `ParticleCollection`. For sure I would keep the `asdict` a **function** (just like `attr.asdict`). The question then is whether `fromdict` should be a static **method** of the class itself or a **function** as well (as it's implemented now).

spflueger: Ok, I will just drop that design discussion here, since this will take quite a while to nail down, preventing us to wrap up the release. 

So regard the following more as a comment:
In my opinion, the attr.asdict design works because of the whole attr structure. That every attr class defines fields (attributes) and their corresponding types and values. Since a normal class does not provide that a special method has to be written for every class. Then I don't see the same design is enforced here.
So either we change the class into some plain data structure, hence attr can be used. Otherwise an interface should be defined and these functions should become the implementations of that interface. I think this will also solve this circular dependency problem that appeared 
> load_default_particles had to be moved to the top import, so that the reaction module is not dependent on the io module (the type hint for Result in the io introduces a circular dependency).

redeboer: Yes, this issue is related to #325 as well I would say, so I made a note there already.

redeboer: I'm reformulating and reopening #383. @spflueger could you list the pros and cons there?

src/expertsystem/reaction/__init__.py
redeboer: So the default has changed slightly: (only) this particle is not provided anymore by default
https://github.com/ComPWA/expertsystem/blob/1e8b81773342b022cdfe41d18d8d6e4c6e540dcf/src/expertsystem/particle/additional_definitions.yml#L2-L13

spflueger: Ok

src/expertsystem/io/_dict.py
spflueger: Ok done

spflueger: Then lets move the design discussion there

spflueger (approved):

PR #531, opened 2021-03-24
refactor: rename parameters to parameter_defaults

redeboer:
This naming makes more sense, as `HelicityModel.parameters` sounds like some sequence that lists the parameters in the model, while it's actually a mapping of those parameters to default ('suggested') values.


spflueger:
Sounds reasonable here. Mostly because these parameters are not connected to the model. Only if you do some action like `subs`. I want to point out though, that in the tensorwaves Model interface, the parameters are actually connected to the model. So there the name defaults would be wrong.

Issue #372, opened 2020-11-04
Use the Gellmann Nishijima rule in the Particle initialization


spflueger:
Extracted from #303

PR #551, opened 2021-03-29
refactor!: move particle module to reaction

redeboer:
The `particle` module was originally under a separate module, because it could theoretically operate on its own. However, since its main classes `Particle`, `ParticleCollection` and `Spin` merely serve as user interfaces to the `reaction` module, it's better to move it back into the `reaction` module.

This PR also enables #372, which was currently not possible due to cyclic imports.


redeboer:
`expertsystem` 0.7.2 should be released first, because this is a major change to the interface.

In addition, note that at this stage it starts to make more sense to split the `expertsystem` into some `pwa-reaction` and `pwa-amplitude` package. This can be easily done now that `io` only operates on classes from the `reaction` module.

spflueger (approved):

